PR #37, opened 2025-01-25
Taylor/ope-66-update-mae-metrics-to-limit-potential-forecast-deviation-by

aaTman:
# EWB Pull Request

## Description

To account for forecasts that might have two consecutive maximums, we decided that truncating to +-48 hours around the observed maximum prevents erroneous and unfair comparisons between forecast and observed maximums. This code addresses this enhancement, moves some functions around, cleans up some functions, "bugfixes" the MaxOfMinTempMAE and MaximumMAE to handle the truncation, and adds test coverage to new functions.

We have "hardcoded" the 48 hours into the function argument but can consider a more user-friendly approach to modifying this if desired.

## Type of change

## How Has This Been Tested?

Ran:

1. pytest locally
2. pre-commit locally
3. a sample run through of the entire EWB evaluate pipeline

## Checklist:


aaTman:
There are three parts of metrics.py that is not covered by tests which I am going to address before merging this PR. I want to enforce 100% test coverage as a minimum for metrics. 100% test coverage is a minimum as there are a myriad of edge cases which 100% test coverage does not account for. Obviously we can spend an endless amount of time thinking up and coding those edge cases but we should at least think up, estimate, and push some of the most expected edge cases.

src/extremeweatherbench/metrics.py
```diff
@@ -25,25 +24,37 @@ def name(self) -> str:
...
+        in case the observations are at a higher temporal resolution than forecast data.
+        Metrics which need a singular timestep from gridded obs will fail if the forecasts
+        are not aligned with the observation timestamps (e.g. a 03z minimum temp in observations
+        when the forecast only has 00z and 06z timesteps).
+        Args:
+            forecast: The forecast dataarray to align.
+            observation: The observation dataarray to align.
+        Returns:
+            The aligned observation dataarray.
+        """
```
darothen, line 40: Some nits in how this docstring is formatted:

1. Try to start with a 1-liner that succinctly explains what this method does.
2. Include spaces between paragraphs (text blocks) and the "Args" and "Returns sections." Vertical whitespace = improved readability.
3. Great to include a paragraph explaining how this function works or what it does.
4. Try to avoid redundancy in explaining variable names. E.g. in Args, we already know that "forecast" is a DataArray and it's going to be aligned, so a more helpful entry might be, "forecast: The forecast data against which the `observation` data will be resampled in time."

src/extremeweatherbench/metrics.py
```diff
@@ -25,25 +24,37 @@ def name(self) -> str:
...
-        """Align the forecast and observation datasets."""
-
-        forecast = forecast.sel(init_time=init_time_datetime)
-        time = np.array(
-            [
-                init_time_datetime + pd.Timedelta(hours=int(t))
-                for t in forecast["lead_time"]
-            ]
+    def _align_observations_temporal_resolution(
+        self, forecast: xr.DataArray, observation: xr.DataArray
```
darothen, line 28: nit: I think this method could be called, "temporally_align_observations_with_forecast". Your call.

darothen: Maybe more importantly - why is this an instance method of `Metric`? You never actually refer to the `self` argument, which is a good hint that this is a generic function that might be along in a "utils" library or module.

src/extremeweatherbench/metrics.py
```diff
@@ -25,25 +24,37 @@ def name(self) -> str:
...
+                # Resample observations to match forecast resolution
+                observation = observation.resample(time=forecast_time_delta).first()
+            else:
+                logger.warning(
+                    "Observation time resolution (%s) is coarser than forecast time resolution (%s)",
+                    obs_time_delta,
+                    forecast_time_delta,
+                )
+
+        return observation
```
darothen, line 57: I don't quite understand why this entire method wouldn't just be line 49? Maybe more fundamentally - it seems awkward to try to resample to a time delta, I'd expect to see

``` python
obs_resampled = observation.resample(time=forecast.valid_time).first()
```

or similar?

src/extremeweatherbench/metrics.py
```diff
@@ -53,8 +64,10 @@ class RegionalRMSE(Metric):
     def compute(self, forecast: xr.DataArray, observation: xr.DataArray):
         rmse_values = []
         for init_time in forecast.init_time:
-            init_forecast, subset_observation = self.align_datasets(
-                forecast, observation, pd.Timestamp(init_time.values).to_pydatetime()
+            init_forecast, subset_observation = utils.temporal_align_dataarrays(
```
darothen, line 67: Noting that the way you've packaged `utils.temporal_align_dataarrays` here is how I'd expect to see the previous `_align_observations_temporal_resampled` implemented.

src/extremeweatherbench/metrics.py
```diff
@@ -68,32 +81,42 @@ class MaximumMAE(Metric):
...
-        observation_spatial_mean = observation_spatial_mean.where(
-            observation_spatial_mean.time.dt.hour % 6 == 0, drop=True
-        )
-        forecast_spatial_mean = forecast.mean(["latitude", "longitude"])
-        for init_time in forecast_spatial_mean.init_time:
-            if forecast.name == "air_temperature":
-                max_date = observation_spatial_mean.idxmax("time").values
-                max_value = observation_spatial_mean.sel(time=max_date).values
-                init_forecast_spatial_mean, _ = self.align_datasets(
+        if forecast.name == "air_temperature":
```
darothen, line 84: I _really_ dislike the idea that these methods know anything about the contents of either the `forecast` or `observation` DataArrays. The only thing they should care about is dimensionality/dimension names - which we're forcing to a standard by contract. This method shouldn't care if the data is `air_temperature` or `average_velocity_of_unladen_swallow_passing_through_gridcell`. We can't prevent users from doing silly things, so if they abuse a method like this then it's caveat emptor.

src/extremeweatherbench/metrics.py
```diff
@@ -68,32 +81,42 @@ class MaximumMAE(Metric):
...
 
-                if max_date in init_forecast_spatial_mean.time.values:
-                    lead_time = init_forecast_spatial_mean.where(
-                        init_forecast_spatial_mean.time == max_date, drop=True
+                if max_datetime in init_forecast_spatial_mean.time.values:
+                    # Subset to +-48 hours centered on the maximum temperature timestamp
+                    filtered_forecast = utils.center_forecast_on_time(
+                        init_forecast_spatial_mean,
+                        time=pd.Timestamp(max_datetime),
+                        hours=48,
```
darothen, line 105: Seems like `hours` is a good candidate for a parameter that `RegionalRMSE` should accept.

src/extremeweatherbench/metrics.py
```diff
@@ -103,73 +126,108 @@ class MaxOfMinTempMAE(Metric):
...
-                        .min()
-                        .max()
-                    ),
-                    drop=True,
-                ).time
-                max_min_value = observation_spatial_mean.sel(
-                    time=max_min_timestamp
-                ).values
-                init_forecast_spatial_mean, _ = self.align_datasets(
+        if forecast.name == "air_temperature":
```
darothen, line 129: Same comment as above re: filtering based on the `forecast.name`. Just get rid of it; let users be responsible for using these methods appropriately.

src/extremeweatherbench/metrics.py
```diff
@@ -103,73 +126,108 @@ class MaxOfMinTempMAE(Metric):
...
+        # Only keep days that have a full set of timestamps
+        da = da.where(
+            da.time.dt.dayofyear.isin(
+                valid_days.where(valid_days).dropna(dim="dayofyear").dayofyear
+            ),
+            drop=True,
+        )
+        return da
+
+    def _return_max_min_timestamp(self, da: xr.DataArray) -> pd.Timestamp:
```
darothen, line 210: These are two more methods that probably shouldn't be instance methods of `Metrics` or a sub-class; they can probably move to your `utils` library, since we never need to reference `self`.

src/extremeweatherbench/utils.py
```diff
@@ -274,16 +275,11 @@ def expand_lead_times_to_6_hourly(
     return dataarray
 
 
-def process_dataarray_for_output(da_list: List[Optional[xr.DataArray]]):
+def process_dataarray_for_output(da_list: List[Optional[xr.DataArray]]) -> xr.DataArray:
```
darothen, line 278: I'm a little bit surprised that this type hint doesn't throw an error... a `List[Optional[T]]` doesn't really make sense because, are you trying to capture the edge case where the list is empty? An empty list still satisfies `List[T]`, so you're covered there; so easiest just change `da_list: List[xr.DataArray]` and handle the edge case where it's empty explicitly.

src/extremeweatherbench/utils.py
```diff
@@ -274,16 +275,11 @@ def expand_lead_times_to_6_hourly(
...
     """Extract and format data from a list of DataArrays.
 
     Args:
         dataarray: A list of xarray DataArrays.
```
darothen, line 282: Looks like you changed the name of the argument `da_list`.

src/extremeweatherbench/utils.py
```diff
@@ -303,3 +299,40 @@ def process_dataarray_for_output(da_list: List[Optional[xr.DataArray]]):
         output_da = output_da.isel(lead_time=slice(None, None, -1))
     output_da = expand_lead_times_to_6_hourly(output_da)
     return output_da
+
+
+def center_forecast_on_time(da: xr.DataArray, time: pd.Timestamp, hours: int):
+    """Center a forecast DataArray on a given time with a given range in hours.
+    Args:
```
darothen, line 306:
```suggestion
def center_forecast_on_time(da: xr.DataArray, time: pd.Timestamp, hours: int):
    """Center a forecast DataArray on a given time with a given range in hours.
    
    Args:
```

src/extremeweatherbench/utils.py
```diff
@@ -303,3 +299,40 @@ def process_dataarray_for_output(da_list: List[Optional[xr.DataArray]]):
...
+        da: The forecast DataArray to center.
+        time: The time to center the forecast on.
+        hours: The number of hours to include in the centered forecast.
+    """
+    time_range = pd.date_range(
+        end=pd.to_datetime(time) + pd.Timedelta(hours=48), periods=97, freq="h"
+    )
```
darothen, line 313: You accept an `hours` argument but hard-code **hours** and **periods** here... should generalize to use the arguments from the user.

src/extremeweatherbench/utils.py
```diff
@@ -303,3 +299,40 @@ def process_dataarray_for_output(da_list: List[Optional[xr.DataArray]]):
...
+    init_time_datetime: datetime.datetime,
+) -> tuple[xr.DataArray, xr.DataArray]:
+    """Align the individual initialization time forecast and observation dataarrays.
+    Args:
+        forecast: The forecast dataarray to align.
+        observation: The observation dataarray to align.
+        init_time_datetime: The initialization time to subset the forecast dataarray by.
+    Returns:
+        A tuple containing the time-aligned forecast and observation dataarrays.
+    """
```
darothen, line 329:
```suggestion
    """Align the individual initialization time forecast and observation dataarrays.
    
    Args:
        forecast: The forecast dataarray to align.
        observation: The observation dataarray to align.
        init_time_datetime: The initialization time to subset the forecast dataarray by.

    Returns:
        A tuple containing the time-aligned forecast and observation dataarrays.
    """
```

tests/conftest.py
```diff
@@ -57,8 +75,8 @@ def mock_config():
 @pytest.fixture
 def mock_gridded_obs_dataset():
```
darothen, line 76: I'd call this "sample_gridded_ob_dataset"... "mock" means something else (you'd typically mock interfaces and actions, but this is very much just sample data).

tests/conftest.py
```diff
@@ -69,31 +87,34 @@ def mock_gridded_obs_dataset():
...
-    dataset = xr.Dataset(
-        {
-            "2m_temperature": (["time", "latitude", "longitude"], 20 + 5 * data),
-            "tp": (["time", "latitude", "longitude"], data),
-            "10m_u_component_of_wind": (["time", "latitude", "longitude"], data),
-            "10m_v_component_of_wind": (["time", "latitude", "longitude"], data),
-        },
-        coords={"time": time, "latitude": latitudes, "longitude": longitudes},
+def mock_forecast_dataarray(mock_forecast_dataset):
+    return dataset_to_dataarray(mock_forecast_dataset)
```
darothen, line 103: I'm not sure if it's kosher to compose fixtures like this... I generally avoid it.

I would recommend defining the actual `xr.Dataset` in a stand-alone method, "make_sample_forecast_dataset() -> xr.Dataset". Then, your fixture `sample_forecast_dataset` can just call this function and return it naked, and this "sample_forecast_dataarray" fixture can call it and return "dataset_to_datarray" applied to it - no need to pass in the dataset fixture, which will make life in the actual unit test implementation more managable.

tests/test_metrics.py
```diff
@@ -27,25 +27,50 @@ def test_compute_not_implemented(
         with pytest.raises(NotImplementedError):
             metric.compute(mock_forecast_dataarray, mock_gridded_obs_dataarray)
 
-    def test_align_datasets(self, mock_forecast_dataarray, mock_gridded_obs_dataarray):
-        """Test that the conversion from init time to valid time (named as time) produces an aligned
-        dataarray to ensure metrics are applied properly."""
+    """Tests the _align_observations_temporal_resolution method of the Metric base class."""
+
```
darothen, line 31: You don't need this doc string, the examples are self-explanatory. If it were very important to split out the tests for this functionality, you'd make a new `class` that only tests these functions, e.g. `class TestMetricObsAlignment` or something.

tests/test_metrics.py
```diff
@@ -71,47 +96,47 @@ def test_regional_rmse_values(
         to a reasonable (1e-7) precision."""
         output = np.array(
```
darothen, line 97: I would strongly recommend saving this and other static assets in a .npz file and implementing a utility to load the data.

darothen (changes requested):
Nice comprehensive set of updates here. Have a fair deal of feedback for us to work through before merging.

src/extremeweatherbench/metrics.py
aaTman: I was thinking about this earlier. Primarily, I don't see it being called outside of the Metrics subclasses, so I kept it within the Metrics parent class from an organizational perspective. I had considered it being a generic function to put into utils, happy to make that change.

src/extremeweatherbench/metrics.py
aaTman: on it. all valid and good suggestions.

src/extremeweatherbench/metrics.py
aaTman: absolutely valid, though i now hope someone tries this with avian data.

src/extremeweatherbench/metrics.py
aaTman: One of the reasons I had set these up in the class was thinking about how to establish a solid, modular paradigm for users to integrate their own metrics into the library. Hence, adding the helper functions into the class itself despite not needing to reference self. I can move these over to `utils`, but wanted to get your thoughts on that before I did.

src/extremeweatherbench/utils.py
aaTman: Thanks for the tip! :D

src/extremeweatherbench/metrics.py
aaTman: Because forecast is using the init_time + lead_time paradigm, it avoids adding a step of converting to valid time to align with observations. This approach simplifies to up/downsample the data as opposed to multiple-step alignment. That said, I can definitely concoct something to avoid timedelta comparisons if you think it's too awkward.

src/extremeweatherbench/utils.py
```diff
@@ -303,3 +298,96 @@ def process_dataarray_for_output(da_list: List[Optional[xr.DataArray]]):
...
+def center_forecast_on_time(da: xr.DataArray, time: pd.Timestamp, hours: int):
+    """Center a forecast DataArray on a given time with a given range in hours.
+
+    Args:
+        da: The forecast DataArray to center.
+        time: The time to center the forecast on.
+        hours: The number of hours to include in the centered forecast.
+    """
+    time_range = pd.date_range(
+        end=pd.to_datetime(time) + pd.Timedelta(hours=hours),
```
darothen, line 312: I was surprised this works :) nifty trick to create a date range ending on a precise timestamp. Cool!

darothen (approved):
Looks great, thanks! I appreciate the great detail here, especially with all the test cases. They will pay dividends moving forward...

PR #42, opened 2025-01-31
Amy/ope 69 case information publicized on repository

amymcgovern:
# EWB Pull Request

## Description

This adds all of the documentation of the EWB case studies as we had previously collected.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

This is only a documentation update (though a big one) so there is no testing performed.

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings


amymcgovern:
@darothen Tagging

darothen:
It worked! I'll check this out as soon as I can.

amymcgovern:
Tagging @gideonite

gideonite (approved):
I see a lot of documentation here. I haven't checked any of it and I don't understand its contents. It looks like you are looking for a rubber stamp approval, which I am happy to provide.

A few broad points of feedback:
1. I don't see any "link tests" if a file name changes or if a file is removed, links will break. This makes the documentation brittle
2. Markdown tables are difficult to edit. I'm not sure what the long term vision is for these tables, but it should take that into account.
3. I don't see many references to the code here. This leads me to think that this documentation should be in a different format / location.

It seems like this should get merged but I would recommend re-thinking how we track this (seemingly valuable) work and information, but that can come in a follow-up PR / change.

PR #44, opened 2025-02-04
Taylor/ope-71-running-ewb-when-installed-via-conda-fails-due-to-path-issue

aaTman:
# EWB Pull Request

## Description

The massive number of file changes here are almost entirely due to a change in directory structure. Based on [this link](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files), it seems appropriate to move all data into the src/extremeweatherbench directory and include it in the package install (as it is more or less required for the library to function). I also modified the evaluate code to use pathlib instead of os.

I also have two small changes in the metrics code to unbreak the pytests. This exists in a PR also ready to merge.

## Type of change

## How Has This Been Tested?

pytest and pre-commit were run, as well as a dummy run through on a new environment. To note: something weird is happening with the MLWP parquets. That will be a tomorrow problem.

## Checklist:


src/extremeweatherbench/evaluate.py
```diff
@@ -36,9 +36,8 @@ def evaluate(
         A dictionary mapping event types to lists of xarray Datasets containing the
         evaluation results for each case within the event type.
     """
-
-    base_dir = os.path.dirname(os.path.abspath(__file__))
-    events_file_path = os.path.join(base_dir, "../../assets/data/events.yaml")
+    base_dir = Path(__file__).resolve().parent
+    events_file_path = base_dir / "assets" / "data" / "events.yaml"
```
darothen, line 40: Generally we don't want to hard-code paths like this; Python provides tools to package and ship data with distributions, and we want to leverage that machinery. You want to look up [importlib](https://docs.python.org/3/library/importlib.html) and follow through with how to register datasets in your package via the **pyproject.toml** listings.

You can see a worked example of how this winds up implemented in code [here](https://github.com/brightbandtech/debufr/blob/cb5b910353e20aaa4daa0ffa7afc81fdb2605989/src/debufr/code_flag.py#L163).

darothen (changes requested):
Requested a change to use `importlib` to manage the default assets you ship with your package.

Another quick note - this is fine for now, but we generally want to avoid shipping datasets beyond very small, single-file assets within the actually library. There's a standard tool called [git-lfs](https://git-lfs.com/) which is commonly used for this - you basically tell it to track files in a repo, and it will manage things for you. It's an easier way to version control and ship static data assets.

darothen (approved):
LGTM

PR #46, opened 2025-02-04
Taylor/ope-67-point-sources-generated-for-each-case

aaTman:
# EWB Pull Request

## Description

Added a parquet file for all of the cases up to this point with ISD data that fit within the bounding box of each case.

A few caveats:

1. WeatherReal code was used to do some quality control. The final step in their 4 step process (quality_control) was not performed. 

2. Areas with sparse or non-existant data (e.g. Afghanistan) produce few ISD data sources, or none at all.

3. Some ISD stations produced erratic, inconsistent data such as in India.

Addressing 2 and 3 will be the subject of a future PR.

Optional dependencies were added to dev for users who want to generate the data themselves, which is currently being used as a catch-all for anything not in the core repo. 

## Type of change

## How Has This Been Tested?

pytest and pre-commit hooks were run locally. New file with code to generate dataframe was also tested locally.

## Checklist:


src/extremeweatherbench/utils.py
```diff
@@ -63,24 +63,21 @@ def generate_json_from_nc(u, so, fs, fs_out, json_dir):
...
-def clip_dataset_to_bounding_box(
-    dataset: xr.Dataset,
-    location_center: Location,
-    length_km: float,
-) -> xr.Dataset:
-    """Clip an xarray dataset to a boxbox around a given location.
+def get_bounding_corners(
+    location_center: Location, length_km: float, convert_to_360: bool = True
+) -> tuple:
+    """Find corners of bounding box given location center and box size in kilometers.
```
darothen, line 69: On re-reading this, I think it might be better to just request the box size in degrees; the width of a degree longitude depends on what your latitude is. In L83 you use a fixed constant which works okay, but I can imagine scenarios where this is a tad bit too imprecise. If you tweak the `length_km` to `length_degrees` it should be totally fine. You can always add a helper function that accepts length as kilometers, estimates that as degrees, and then calls this function.

src/extremeweatherbench/utils.py
```diff
@@ -90,7 +87,30 @@ def clip_dataset_to_bounding_box(
...
+    if convert_to_360:
+        min_lon = convert_longitude_to_360(min_lon)
+        max_lon = convert_longitude_to_360(max_lon)
+    return (min_lat, max_lat, min_lon, max_lon)
+
 
+def clip_dataset_to_bounding_box(
+    dataset: xr.Dataset,
+    location_center: Location,
+    length_km: float,
```
darothen, line 99: Another thing to think about here... since this is named **clip_dataset_to_bounding_box**, I'd expect that a bounding box would be an argument - especially since you have another method which allows users to create bounding boxes!

The trade-off of not calling out to **get_bounding_corners** here is really just that users have to write two lines of code instead of one:

``` python
# Original
loc = Location(0, 0)
new_ds = clip_dataset_to_bounding_box(orig_ds, loc, 1.0)
# New
loc = Location(0, 0)
bbox = get_bounding_corners(loc, 1.0)
new_ds = clip_dataset_to_bounding_box(orig_ds, bbox)
```

I think the "new" code is a bit easier to understand. "Explicit is better than implicit."

darothen (changes requested):

PR #61, opened 2025-02-14
Taylor/ope-51-point-observations-included-in-ewb

aaTman:
# EWB Pull Request

## Description

This PR adds in point observation functionality in a MWE. There's some significant streamlining that should occur immediately following this PR, but wanted to keep the ball rolling and get this reviewed + pushed.

Key points:

1. Point observations, using (for now) a pre-computed parquet on the Brightband public GCS bucket, are now enabled as validation.

2. The ISD downloader code ended up with a bug that was fixed when changing to degree-based bounding box code.

3. Some metrics were unable to handle the new point observation dataarrays. This was addressed while maintaining equivalent functionality for gridded observation dataarrays.

4. Due to the dependency of the parquet file, the variable names that forecast data are mapped to were changed to be more explicit (as the current point obs are all surface-based). This will be one less conflict when we start introducing variables that are both surface and level-based, e.g. u/v winds.

5. Some small tweaks on where dask enabled xarray datasets are computed, with an emphasis on the new chunking paradigm in CIRA MLWP.

6. Added a cache directory constant for both handling the stored BallTree data as well as soon-to-be caching stores for users to pick up where they left off.

7. Added the pick_points function from Herbie, with some upgrades and tweaks to synergize with the current EWB code.

8. Tests updated to account for new code.

## Type of change

## How Has This Been Tested?

Pytest + pre-commit + running a local evaluate.evaluate in a jupyter notebook with sample CIRA MLWP data. 

## Checklist:


src/extremeweatherbench/config.py
```diff
@@ -24,36 +26,54 @@ class Config:
...
     """A mapping between standard variable names used across EWB, and their counterpart
     in a forecast dataset.
 
-    Allows users to insert custom schemas for decoding forecast data. Defaults are
-    suggested based on the CF Conventions.
+    Allows users to insert custom schemas for decoding forecast data. Defaults mappings are
+    based on the CF Conventions. Default values are based on the CIRA MLWP archive.
+
+    Attributes:
+        surface_air_temperature: The name of the variable representing surface air
```
aaTman, line 58: I feel like I'm being a little _too_ explicit here, can easily cut this docstring down

src/extremeweatherbench/evaluate.py
```diff
@@ -164,7 +156,9 @@ def _evaluate_case(
 
     logger.info("Evaluating case %s, %s", individual_case.id, individual_case.title)
     variable_subset_ds = individual_case._subset_data_vars(forecast_dataset)
-    time_subset_forecast_ds = individual_case._subset_valid_times(variable_subset_ds)
+    time_subset_forecast_ds = (
+        individual_case._subset_valid_times(variable_subset_ds).compute()
+    )  # based on a (1,1,721,1440) chunk size this is where compute is more efficient
```
darothen, line 161: I don't quite understand this comment... what is more efficient?

src/extremeweatherbench/evaluate.py
```diff
@@ -194,25 +192,112 @@ def _evaluate_case(
...
+            )
+            & (
+                case_subset_point_obs["longitude"]
+                >= spatiotemporal_subset_forecast_ds["longitude"].min().values
+            )
+            & (
+                case_subset_point_obs["longitude"]
+                <= spatiotemporal_subset_forecast_ds["longitude"].max().values
+            )
+        ]
```
darothen, line 233: This should probably be a helper function with independent testing (lots of "gotchas" can arise on the edges of this sort of interval selection).

src/extremeweatherbench/utils.py
```diff
@@ -414,3 +413,338 @@ def read_event_yaml(input_pth: str | Path) -> dict:
     with open(input_pth, "rb") as f:
         yaml_event_case = yaml.safe_load(f)
     return yaml_event_case
+
+
+# BallTree algorithm from Herbie
+# (https://github.com/blaylockbk/Herbie/blob/a96cbf5ea864c9a85975bf98fa90a56381c9aa75/herbie/accessors.py#L310)
+def pick_points(
```
darothen, line 420: We can't lift code verbatim from a third-party source, even if it has a permissive license. Even with attribution, this has a lot of risk - at the most basic level, if there are bugs in this routine that aren't obvious and get fixed upstream, we have no way to know that or benefit from those improvements.

This should be removed and if it's 100% vital and necessary, then we have to add Herbie as a dependency to this project. That is also probably not the best idea, since Herbie is a super heavy package since it has direct imports to compiled tools like `proj` and requires installation via `conda`. 

So we have two options to choose from:

1. Write a *minimal working version* of the algorithm here. It's appropriate - and encouraged! - to directly cite the code from Herbie as the design template and reference implementation. But we should write a version of the code that implements _only_ the functionality we need, and nothing more. 

2. Pursue a different library or methodology.

We should review these options live, will reach out and put something on your calendar.

src/extremeweatherbench/evaluate.py
```diff
@@ -194,25 +192,112 @@ def _evaluate_case(
...
                 result = metric_instance.compute(forecast_da, gridded_obs_da)
-                case_results[data_var][metric_instance.name] = result
+                case_results["gridded"][data_var][metric_instance.name] = result
+                logger.debug(
+                    "gridded, %s, %s, %s", data_var, metric_instance.name, result
+                )
     if point_obs is not None:
-        raise NotImplementedError("Point obs evaluation not implemented as of 0.1.0")
+        case_results["point"] = {}
+        case_subset_point_obs = point_obs.loc[point_obs["id"] == individual_case.id]
```
darothen, line 210: What is "id" here and what does it mean to be equivalent between the point observation dataset and the case specification? Does this imply that observations can only be used for a single case?

src/extremeweatherbench/evaluate.py
```diff
@@ -194,25 +192,112 @@ def _evaluate_case(
...
+
+        point_forecast_subset_ds = utils.pick_points(
+            spatiotemporal_subset_forecast_ds,
+            case_subset_point_obs.drop_duplicates(
+                subset=["latitude", "longitude", "station"]
+            )[["latitude", "longitude", "elev", "id", "name"]],
+            config=eval_config,
+            tree_name=individual_case.id,
+        )
+        point_forecast_subset_ds = point_forecast_subset_ds.rename({"point": "station"})
```
darothen, line 256: This call is doing a tremendous amount of heavy lifting. You should add comments to explain what is happening; at a minimum, I think you need a paragraph-length comment preceding this block explaining the format of the data sources you're comparing and what the expected output will look like.

src/extremeweatherbench/evaluate.py
```diff
@@ -252,7 +337,9 @@ def _open_obs_datasets(eval_config: config.Config):
     point_obs = None
     gridded_obs = None
     if eval_config.point_obs_path:
-        raise NotImplementedError("Point obs evaluation not implemented as of 0.1.0")
+        point_obs = pd.read_parquet(
+            eval_config.point_obs_path, storage_options=dict(token="anon")
+        )
```
darothen, line 342: `storage_options` should probably be lifted to a module-level parameter that you can define as a sane default, with the explanation that we provide access to a curated set of observations gratis for users.

src/extremeweatherbench/evaluate.py
```diff
@@ -194,25 +192,112 @@ def _evaluate_case(
...
+                    "point, %s, %s, %s", data_var, metric_instance.name, result
+                )
     return case_results
 
 
 def _open_forecast_dataset(
     eval_config: config.Config,
     forecast_schema_config: config.ForecastSchemaConfig = DEFAULT_FORECAST_SCHEMA_CONFIG,
 ):
+    # TODO: #57 make all this much faster and ...better code
```
darothen, line 300: Similar comment as before, please add enough information here that we can understand what is slow about the code, why it needs to be faster, and why these improvements aren't yet implemented.

src/extremeweatherbench/metrics.py
```diff
@@ -135,31 +146,42 @@ def compute(self, forecast: xr.DataArray, observation: xr.DataArray):
...
+                        filtered_forecast
+                        == filtered_forecast.groupby("time.dayofyear")
+                        .min()
+                        .max(),  # TODO check if "time" "dayofyear" as it is in point obs works here
+                        drop=True,
+                    )
+
+                    # TODO: add temporal displacement error, which is
+                    # filtered_forecast_max_min.time.values[0] - max_min_timestamp
+                    if max_min_timestamp in filtered_forecast.time.values:
```
darothen, line 171: By my count, we're three-levels deep with this `if` statement. What is going on here, and why do we need such deeply nested code? This is usually a sign that we should invert the logic in our if/else's and employ short-circuiting, or re-factor some of this logic wholesale to utility functions. Code this deeply nested is very hard to follow because you have to keep multiple, prior conditionals and their context in mind, which is quite difficult in practice.

darothen (changes requested):
I left a lot of actionable comments, but there are a few high-level bits of feedback:

1. This needs to be broken down into multiple PRs - at least 2, although you could easily justify 3+. At a minimum, the logic surrounding this merger between observations/forecasts on lat/lons needs an exclusive commit because it's extremely complex and has many lines of code. A tell-tale sign that a PR is too big is if the diff for a file doesn't even render in GitHub by default.

2. We need to separately discuss the alignment algorithm. See my comment in `utils.py` about copy/pasting from the Herbie package. I would recommend that we make a stand-alone Jupyter notebook that works an annotated example so we can discuss each of the steps and understand why we need this `pick_points` function. I **really** do not want to use that function because it has heavyweight dependencies which will make this package _significantly_ more difficult for users to install and use.

3. If you're ever leaving a "TODO" or note about a deficiency in the code, it **must** be accompanied with supporting information. The motivation here is that it could be days, months, or years before you ever revisit that code, and it's unreasonable to expect you'll recall all the context around that note. For our internal development, I restrict the options here to one of two things: (a) file a proper Issue in GitHub and fully document the problem that needs to be resolved, then write your in-code comment in such a way that it gets auto-linked in GitHub; or (b) write a NOTE that spends a few sentences explaining what you did, what is wrong with it, and any inklings you have about a better solution.

I know there's a lot here, so let's find time on Monday or Tuesday to chat and break this down into more actionable follow-up work.

src/extremeweatherbench/case.py
```diff
@@ -133,17 +133,18 @@ class IndividualHeatWaveCase(IndividualCase):
             metrics.MaximumMAE,
         ]
     )
+    # TODO: remove default factory?; simplify to just data_vars
     data_vars: List[str] = dataclasses.field(
```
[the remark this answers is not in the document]
aaTman: Definitely a note, this was something found unrelated to the PR that should be addressed.

src/extremeweatherbench/case.py
```diff
@@ -133,17 +133,18 @@ class IndividualHeatWaveCase(IndividualCase):
...
-        default_factory=lambda: ["air_temperature"]
+        default_factory=lambda: ["surface_air_temperature"]
```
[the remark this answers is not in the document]
aaTman: Agreed, I think there was something funky occurring with the default factory. This is related to the TODO comment you made above, the only thing changed in this PR was the name of the data variable.

src/extremeweatherbench/case.py
```diff
@@ -133,17 +133,18 @@ class IndividualHeatWaveCase(IndividualCase):
...
 
     def __post_init__(self):
-        self.data_vars = ["air_temperature"]
+        self.data_vars = ["surface_air_temperature"]
 
     def perform_subsetting_procedure(self, dataset: xr.Dataset) -> xr.Dataset:
         modified_ds = utils.clip_dataset_to_bounding_box_degrees(
             dataset, self.location, self.bounding_box_degrees
-        )
+        ).compute()
```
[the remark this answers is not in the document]
aaTman: I can do some tests but found some interesting time to compute disparities between using `.compute()` early on during the subsetting versus later. This was the location that produced the fastest return for a few datasets tested, including zarr and kerchunk references that had a chunksize of (1,1,721,1440).

src/extremeweatherbench/config.py
```diff
@@ -5,11 +5,13 @@
 
 from extremeweatherbench import events
 
-DEFAULT_OUTPUT_DIR = "assets/data/outputs/"
-DEFAULT_FORECAST_DIR = "assets/data/forecasts/"
+DEFAULT_OUTPUT_DIR = "data/outputs/"
+DEFAULT_FORECAST_DIR = "data/forecasts/"
+DEFAULT_CACHE_DIR = "data/cache/"
```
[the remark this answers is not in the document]
aaTman: I like this a lot better. I'll work on integrating it into my workflow moving forward.

src/extremeweatherbench/config.py
```diff
@@ -24,36 +26,54 @@ class Config:
...
         event_types: A list of event types to evaluate.
-        cache: Enable caching of intermedaite data and artifacts generated by an EWB
+        cache: Enable caching of intermediate data and artifacts generated by an EWB
             analysis. Defaults to "False".
     """
 
     event_types: List[events.EventContainer]
     output_dir: str = DEFAULT_OUTPUT_DIR
     forecast_dir: str = DEFAULT_FORECAST_DIR
+    cache_dir: Optional[str] = DEFAULT_CACHE_DIR
```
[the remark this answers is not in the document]
aaTman: My thought process here was a user could pass `None` explicitly, to avoid caching entirely in the scenario they might be on a restricted permission machine. Or, in the case they simply didn't want to cache.

src/extremeweatherbench/evaluate.py
```diff
@@ -109,6 +101,7 @@ def evaluate(
                 if x
             ]
         ),
+        # TODO: This does not work properly, needs to skip None values
         sum(len(results) for results in all_results.values() if results is not None),
```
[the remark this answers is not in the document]
aaTman: It returns a sum of all cases processed, regardless of if they returned `None` or not. In the test cases I had set up, there are ~20 cases that are skipped due to missing data in the forecast dataset. It should only return the number of cases that processed successfully.

src/extremeweatherbench/evaluate.py
```diff
@@ -119,6 +112,7 @@ def _evaluate_cases_loop(
     forecast_dataset: xr.Dataset,
     gridded_obs: Optional[xr.Dataset] = None,
     point_obs: Optional[pd.DataFrame] = None,
+    eval_config: config.Config = None,
```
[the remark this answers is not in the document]
aaTman: Wonder how mypy didn't catch this. Thanks!

src/extremeweatherbench/evaluate.py
aaTman: `id` in this case is the unique ID number for each case. The observations, on a row by row bases, are exclusive to a case. In the situation with two cases that have overlapping dates, a second row with identical observations (but different `id`) will exist. Because this is our curated parquet data, we haven't experience an overlapping event but might. It's in the wheelhouse to modify that column to be able to hold multiple ID's, preventing duplicates.

src/extremeweatherbench/evaluate.py
aaTman: Definitely, this is part of the next PR which is essentially ready to go that offloads data loading to another file entirely.

src/extremeweatherbench/config.py
```diff
@@ -24,36 +26,54 @@ class Config:
...
     gridded_obs_path: str = ARCO_ERA5_FULL_URI
-    point_obs_path: Optional[str] = None
-    cache: bool = False
+    point_obs_path: str = ISD_POINT_OBS_URI
 
 
 @dataclasses.dataclass
 class ForecastSchemaConfig:
     """A mapping between standard variable names used across EWB, and their counterpart
     in a forecast dataset.
```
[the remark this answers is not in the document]
aaTman: Primarily that it is immutable as a store and a source of metadata but also a useful source to store code that involves changes to the forecast data. One example is in my forthcoming PR, I included a function `conversions(dataset)` that can store simple or more complex changes to incoming data's variables:

```python
@dataclasses.dataclass
class ForecastSchemaConfig:
    """A mapping between standard variable names used across EWB, and their counterpart
    in a forecast dataset.

    Allows users to insert custom schemas for decoding forecast data. Defaults mappings are
    based on the CF Conventions. Default values are based on the CIRA MLWP archive.

    Attributes:
        surface_air_temperature: The name of the variable representing surface air
            temperature.
        surface_eastward_wind: The name of the variable representing eastward wind
            at 10m.
        surface_northward_wind: The name of the variable representing northward wind
            at 10m.
        air_pressure_at_mean_sea_level: The name of the variable representing air
            pressure at mean sea level.
        lead_time: The name of the dimension representing lead time.
        init_time: The name of the dimension representing initialization time.
        fhour: The name of the dimension representing forecast hour.
        level: The name of the dimension representing vertical level.
        latitude: The name of the dimension representing latitude.
        longitude: The name of the dimension representing longitude.
    """

    surface_air_temperature: Optional[str] = "t2"
    surface_eastward_wind: Optional[str] = "u10"
    surface_northward_wind: Optional[str] = "v10"
    air_pressure_at_mean_sea_level: Optional[str] = "msl"
    lead_time: Optional[str] = "time"
    init_time: Optional[str] = "init_time"
    level: Optional[str] = "level"
    latitude: Optional[str] = "latitude"
    longitude: Optional[str] = "longitude"

    def to_mapping(self) -> dict:
        return {
            getattr(self, field.name): field.name for field in dataclasses.fields(self)
        }

    def convert_lead_time(self, dataset: xr.Dataset) -> xr.Dataset:
        """
        Convert types of variables in an xarray Dataset based on the schema,
        ensuring that, for example, the variable representing lead_time is of type int.

        Args:
            dataset: The input xarray Dataset that uses the schema's variable names.

        Returns:
            An xarray Dataset with adjusted types.
        """

        var = dataset["lead_time"]
        if var.dtype == np.dtype("timedelta64[ns]"):
            dataset["lead_time"] = (var / np.timedelta64(1, "h")).astype(int)
        return dataset

    def conversions(self, dataset: xr.Dataset) -> xr.Dataset:
        """Apply all conversions defined in the schema to the dataset."""
        if "lead_time" in dataset.coords:
            dataset = self.convert_lead_time(dataset)
        else:
            logger.warning("No lead_time dimension found in dataset")
        return dataset
```

In this case, `conversions(dataset)` checks for whether or not `lead_time` exists, and if so, if it's `timedelta64[ns]`, like the WeatherBench zarrs. It's one thing to map the variables by name, it's another to manage conversions which I felt was necessary and relevant to the coming PR that involves data loading.

src/extremeweatherbench/case.py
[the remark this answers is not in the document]
aaTman: Going to remove it for now.

src/extremeweatherbench/evaluate.py
aaTman: When testing on forecast data that with (1,1,721,1440) (time, level, lat, lon) chunks, eagerly loading into memory here was much faster on both my local and via the GC VM than anywhere else, and might be a candidate for where to cache intermediate results.

aaTman:
Closing this PR in favor of a cleaner PR soon.

PR #66, opened 2025-02-21
Taylor/ope-88-add-core-point-subsetting-logic

aaTman:
# EWB Pull Request

## Description

This PR adds the core function that subset the grid in the forecast where the points obs are closest to, as well as the location in evaluate.py for the function to be invoked.

## Type of change

## How Has This Been Tested?

This currently "does not work" but will work along with other PR's that will be introduced.

## Checklist:

- [] Any dependent changes have been merged and published in downstream modules


src/extremeweatherbench/utils.py
```diff
@@ -414,3 +431,175 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+
+    Returns a corrected dataframe."""
+    if "surface_air_temperature" in df.columns:
+        # Check if temperature appears to be in Kelvin (values mostly > 200)
+        if df["surface_air_temperature"].mean() > 200:
+            pass
+        else:
+            # Otherwise assume Celsius
+            df["surface_air_temperature"] = df["surface_air_temperature"] + 273.15
+    return df
```
darothen, line 452: I don't think we need this - it could get endlessly complicated, and even in the temperature case here, the implementation is based purely on assumptions about the distribution of values in a column (which itself has an assumption about the column name).

It's far easier to rely on "garbage in, garbage out." Make explicit the requirements for forecast data and validating observations (e.g. units), and then let _caveat emptor_. It'll be obvious enough in the results.

src/extremeweatherbench/utils.py
```diff
@@ -414,3 +431,175 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+    for init_time, lead_time in itertools.product(
+        forecast_da.init_time, forecast_da.lead_time
+    ):
+        valid_time = pd.Timestamp(
+            init_time.values + pd.to_timedelta(lead_time.values, unit="h").to_numpy()
+        )
+        if valid_time in case_subset_point_obs_df.index.levels[1]:
+            obs_timeslice = case_subset_point_obs_df.loc[
+                pd.IndexSlice[:, valid_time], :
+            ].reset_index()
```
darothen, line 553: On reading this a second time, I don't think we should have this conditional. I think one of two options would be more concise and easier to understand:

1. Pre-filter the `valid_time` values. You can generate them by unpacking the `itertools.product()` generator in L544, and then just use them to index against the time dimension/index level of **case_subset_point_obs_df**. What's left can probably be processed in a vectorized fashion.

2. Re-work the logic in L550 to short-circuit. One approach would be to *invert* the existing conditional:

```
if not (valid_time in case_subset_point_bs_df.index.levels[1]):
    continue
obs_timeslice = ...
# carry on with body of conditional
```

The other approach would be to directly subset the `obs_timeslice` and **continue** the loop if the result is empty. Something like

```
idx = pd.IndexSlice[:, valid_time]
obs_timeslice = case_subset_point_obs.loc[idx, :]
if len(obs_timesliace) < 1:
    continue
obs_timeslice = obs_timeslice.reset_index()
```

The point is to flatten the nested structure of the code to make it more readable. It's also to avoid scanning the index twice (which you do implicitly in the conditional logic and then once again in the actual slicing.

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
 [project.optional-dependencies]
 data-prep = [
     "fsspec>=2024.12.0",
+    "geopy>=2.4.1",
```
darothen, line 29: I know this is an optional dependency, but where do you use this library? I can't find it imported in the code here (maybe my filtered view of the edits is wrong?). If not actually used, then we should remove this (just makes dependency resolution easier).

src/extremeweatherbench/utils.py
```diff
@@ -414,3 +431,175 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+    ].astype("str")
+
+    # Set up multiindex to enable slicing along individual timesteps
+    case_subset_point_obs_df = case_subset_point_obs_df.set_index("time")
+    case_subset_point_obs_df = case_subset_point_obs_df.set_index(
+        "station", append=True
+    )
+    case_subset_point_obs_df = case_subset_point_obs_df.sort_index(
+        level="station"
+    ).swaplevel(0, 1)
```
darothen, line 538: I think can be truncated to one line:

```
case_subset_point_obs_df = case_subset_point_obs_df.set_index(['station', 'time']).sort_index()
```

src/extremeweatherbench/utils.py
```diff
@@ -414,3 +431,175 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+    interpolated_observation = xr.concat(aligned_observation_list, dim="station")
+    return (interpolated_forecast, interpolated_observation)
+
+
+def derive_indices_from_init_time_and_lead_time(
+    dataset: xr.Dataset,
+    start_date: datetime.datetime,
+    end_date: datetime.datetime,
+) -> Tuple[np.ndarray]:
+    if len(dataset.lead_time) == 0 or len(dataset.init_time) == 0:
```
darothen, line 590: Needs docstring; I would definitely encourage showing an example to clarify what is happening here.

src/extremeweatherbench/utils.py
```diff
@@ -414,3 +431,175 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+        raise ValueError("No forecast data available for this case.")
```
darothen, line 591: Are you sure we want to `raise Exception` here? This means that the calling code must now wrap this with a `try/except` block, or else we could randomly crash while executing an analysis if there is some errant data.

The conditional you check in L590 is fully possible within the scope of the calling function. I'd recommend removing this check and putting it there, where you can just decide whether or not to continue processing this `dataset`. It's effectively the same as you'd have to do to properly catch the `ValueError` raised here.

Alternatively, can you just re-work this function to always pass a valid output, even if that output is empty? And then the invoking function or user can decide what to do with empty data. Probably a tad more resilient and easier to cover with testing.

src/extremeweatherbench/utils.py
```diff
@@ -414,3 +431,175 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+    interpolated_forecast = xr.concat(aligned_forecast_list, dim="station")
+    interpolated_observation = xr.concat(aligned_observation_list, dim="station")
+    return (interpolated_forecast, interpolated_observation)
+
+
+def derive_indices_from_init_time_and_lead_time(
+    dataset: xr.Dataset,
+    start_date: datetime.datetime,
+    end_date: datetime.datetime,
+) -> Tuple[np.ndarray]:
```
darothen, line 589: Think about this output argtype - this is saying you'll pass back a `tuple` with a single value, an `np.ndarray`. Is that what we want to pass back? Why wouldn't we just directly pass back the `np.ndarray` in that case?

If you mean that this is a variable length output - that it could be a tuple _of one ore more_ `np.ndarrays`, then there is actually a specific way you indicate that:

```
def func(*args) -> Tuple[np.ndarray,...]:
    pass
```

In either case I'd question why we should return a `tuple` instead of a `list`.

darothen (changes requested):
Commented on some structural changes to your alignment method and some general Python details elsewhere.

pyproject.toml
aaTman: It's in `download_and_process_isd.py`!

src/extremeweatherbench/utils.py
aaTman: it can be, just confirmed that they are equal.

src/extremeweatherbench/utils.py
aaTman: Good point, moved it to `_evaluate_case`!

src/extremeweatherbench/evaluate.py
```diff
@@ -164,6 +164,8 @@ def _evaluate_case(
 
     logger.info("Evaluating case %s, %s", individual_case.id, individual_case.title)
     variable_subset_ds = individual_case._subset_data_vars(forecast_dataset)
+    if len(variable_subset_ds.lead_time) == 0 or len(variable_subset_ds.init_time) == 0:
+        raise ValueError("No forecast data available, check forecast dataset.")
```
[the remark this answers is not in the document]
darothen: Also (including here because it's in existing unmodified code) - I don't like return types of `Optional[...]`; if at all possible, you should have a fixed return type (in this case, `dict[str, xr.Dataset]`) and return an empty dictionary if you hit one of your failure conditions.

Else, name this method `_maybe_evaluate_case` to emphasize that it has some case-dependent outputs.

src/extremeweatherbench/utils.py
```diff
@@ -37,6 +37,17 @@
     "surface_temperature": "surface_air_temperature",
 }
 
+POINT_OBS_METADATA_VARS = [
```
darothen, line 40: Should include a comment similar to above ("#: Struct packaging...") explaining what this is.

src/extremeweatherbench/utils.py
```diff
@@ -133,105 +144,11 @@ def remove_ocean_gridpoints(dataset: xr.Dataset) -> xr.Dataset:
     land_sea_mask = land.mask(dataset.longitude, dataset.latitude)
     land_mask = land_sea_mask == 0
     # Subset the dataset to only include land gridpoints
-    dataset = dataset.where(land_mask, drop=True)
+    dataset = dataset.where(land_mask)
```
darothen, line 147:
```suggestion
    return dataset.where(land_mask)
```

src/extremeweatherbench/utils.py
```diff
@@ -403,3 +320,176 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+        max_lon: maximum longitude.
+
+    Returns a subset dataframe."""
+    location_subset_df = df[
+        (df["latitude"] >= min_lat)
+        & (df["latitude"] <= max_lat)
+        & (df["longitude"] >= min_lon)
+        & (df["longitude"] <= max_lon)
+    ]
+    return location_subset_df
```
darothen, line 348: Equivalent, but slightly cleaner way of writing this code:

```suggestion
    mask = (
        (df["latitude"] >= min_lat) &
        (df["latitude"] <= max_lat) &
        (df["longitude"] >= min_lon) &
        (df["longitude"] <= max_lon)
    )
    return df[mask]
```

This avoids the temporary and makes it very clear what you're computing.

src/extremeweatherbench/utils.py
```diff
@@ -403,3 +320,176 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
     return yaml_event_case
+
+
+def location_subset_point_obs(
+    df: pd.DataFrame,
+    min_lat: float,
+    max_lat: float,
+    min_lon: float,
+    max_lon: float,
+):
```
darothen, line 331: My bet is that you will also want to include two strings with defaults, `..., lat_dim: str = "latitude", lon_dim: str = "longitude"`.

src/extremeweatherbench/utils.py
```diff
@@ -403,3 +320,176 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+    """Takes in a forecast dataarray and point observation dataframe, aligning them by
+    reducing dimensions.
+
+    Args:
+        forecast_ds: The forecast dataset.
+        case_subset_point_obs_df: The point observation dataframe.
+        data_var: The variable to subset (e.g. "surface_air_temperature")
+        point_obs_metadata_vars: The metadata variables to subset (e.g. ["elev", "name"])
+
+    Returns a tuple of aligned forecast and observation dataarrays."""
```
darothen, line 366: End `"""` on its own line.

src/extremeweatherbench/utils.py
```diff
@@ -403,3 +320,176 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+    # Reset index to allow for easier modification
+    case_subset_point_obs_df = case_subset_point_obs_df.reset_index()
+    case_subset_point_obs_df.loc[:, "station"] = case_subset_point_obs_df[
+        "station"
+    ].astype("str")
+
+    # Set up multiindex to enable slicing along individual timesteps
+    case_subset_point_obs_df = case_subset_point_obs_df.set_index(
+        ["station", "time"]
+    ).sort_index()
```
darothen, line 393: This line is a little strange... do you mean to simply create a column named "station" with the values set to the original station but transformed to the `str` type?

If so, I would just do this in one fell swoop with a method chained block:

    
```suggestion 
    case_subset_point_obs_df = (
        case_subset_point_obs_df
        .reset_index()  # Reset index to allow for easier modification
        .assign(station=case_subset_point_obs_df["station"].astype("str")
        .set_index(["station", "time"])  # Set up multiindex to enable slicing along individual timesteps
        .sort_index()
     )
```

src/extremeweatherbench/utils.py
```diff
@@ -403,3 +320,176 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+
+    aligned_forecast_list = []
+    aligned_observation_list = []
+
+    # Loop init and lead times to prevent indexing error from duplicate valid times
+    for init_time, lead_time in itertools.product(
+        forecast_ds.init_time, forecast_ds.lead_time
+    ):
```
darothen, line 401: This would be _great_ place for a `logger.debug` statement indicating how many values there are for **init_time** and **lead_time**, as well as the total number of pairs that will be checked. Likely a place where bugs will creep in with weird user inputs.

src/extremeweatherbench/utils.py
```diff
@@ -403,3 +320,176 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+            init_time.values + pd.to_timedelta(lead_time.values, unit="h").to_numpy()
+        )
+        valid_time_index = pd.IndexSlice[:, valid_time]
+        obs_timeslice = (
+            case_subset_point_obs_df.loc[valid_time_index, :]
+            if valid_time in case_subset_point_obs_df.index.get_level_values(1)
+            else None
+        )
+        if obs_timeslice is None:
+            continue
```
darothen, line 412: Can't you just take `obs_timeslice = case_subset_point_obs_df.loc[valid_time_index, :]` and wrap it in an `IndexError` or expect to take the zero-length DataFrame as the null case, rather than manually handle it by setting to `None`? It's best to use the default or empty value of an output as the base case when possible.

src/extremeweatherbench/utils.py
```diff
@@ -403,3 +320,176 @@ def read_event_yaml(input_pth: str | Path) -> dict:
...
+        )
+        valid_time_subset_obs_ds.coords["lead_time"] = lead_time
+        valid_time_subset_obs_ds.coords["init_time"] = init_time
+
+        aligned_observation_list.append(valid_time_subset_obs_ds)
+        aligned_forecast_list.append(forecast_at_obs_ds)
+    # concat the dataarrays along the station dimension
+    interpolated_forecast = xr.concat(aligned_forecast_list, dim="station")
+    interpolated_observation = xr.concat(aligned_observation_list, dim="station")
+    return (interpolated_forecast, interpolated_observation)
```
darothen, line 441: I think it's okay to keep these outputs like this, because they're logically named and it's easier to read this way.

darothen (approved):
I think we're there; LGTM, with a few comments to quickly clean up before merge.

src/extremeweatherbench/evaluate.py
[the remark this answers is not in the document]
aaTman: The logic here (and this can be changed in the future, of course) is that: if a case is a certain time range (e.g. 2021-01-01 00z to 2021-01-03 00z) and the given forecast data does not have any overlapping valid time (e.g. the forecast data provided is only from 2022-2024), None is returned.

That said, I realized I didn't update naming conventions here from a while ago so thank you for mentioning this. I'll edit the function name too for clarity.

src/extremeweatherbench/utils.py
aaTman: Not a bad idea. I haven't spent time thinking about this but I also don't like the wild-west style of the arguments here. I might do a quick PR to do something more elegant when I have some time for low-priority things. I foresee issues.

src/extremeweatherbench/utils.py
aaTman: Would this suffice? I don't want to break the code in this case, just keep looping: 

```suggestion
try:
    if valid_time in case_subset_point_obs_df.index.get_level_values(1):
        obs_timeslice = case_subset_point_obs_df.loc[valid_time_index, :]
    else:
        continue
```

PR #69, opened 2025-02-24
Taylor/ope-89-update-evaluate-code-to-include-point-obs

aaTman:
# EWB Pull Request

## Description

This adds the MWE logic to integrate point obs into the outputs from EWB.

A forthcoming PR will clean up the duplicated loops on data vars and metrics that are unnecessary. There will also be updated handling of computation - a TODO and issue was created that explains a little bit of the reasoning but that is out of scope for this PR.

## Type of change

## How Has This Been Tested?

pytest, precommit. Once the PRs are all merged into the feature branch, an isolated environment will be used to test dependencies and workflow.

## Checklist:

### DIFF
--- a/src/extremeweatherbench/evaluate.py
+++ b/src/extremeweatherbench/evaluate.py
@@ -186,8 +186,12 @@ def _evaluate_case(
 
     if gridded_obs is not None:
         case_results["gridded"] = {}
-        spatiotemporal_subset_forecast_ds = _subset_gridded_obs(
-            gridded_obs, individual_case
+        variable_subset_gridded_obs = individual_case._subset_data_vars(gridded_obs)
+        time_subset_gridded_obs_ds = variable_subset_gridded_obs.sel(
+            time=slice(individual_case.start_date, individual_case.end_date)
+        )
+        spatiotemporal_subset_forecast_ds = (
+            individual_case.perform_subsetting_procedure(time_subset_gridded_obs_ds)
         )
         # Align gridded_obs and forecast_dataset by time
         subset_gridded_obs, aligned_forecast_subset_variable_time_ds = xr.align(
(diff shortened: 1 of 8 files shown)


src/extremeweatherbench/evaluate.py
```diff
@@ -2,19 +2,183 @@
...
+        self,
+        individual_case: case.IndividualCase,
+        observation_type: Literal["gridded", "point"],
+        forecast: xr.Dataset,
+        observation: Optional[Union[xr.Dataset | pd.DataFrame]],
+    ):
+        self.observation_type = observation_type
+        self.observation = observation
+        self.forecast = forecast
+        self.individual_case = individual_case
```
darothen, line 67: This could (and probably should) also be a `dataclasses.dataclass`. It's basically a wrapper to ingest some datasets and package them together. Then, you can use it as a convenient input to a few functions below (see following comments).

src/extremeweatherbench/evaluate.py
```diff
@@ -2,19 +2,183 @@
...
+            lead_time_len
+            < (self.individual_case.end_date - self.individual_case.start_date).days
+        ):
+            logger.warning(
+                "Fewer valid times in forecast than days in case %s, results likely unreliable",
+                self.individual_case.id,
+            )
+        logger.info("Forecast data available for case %s", self.individual_case.id)
+        return forecast
+
```
darothen, line 181: I don't actually think any of these methods should be bundled as part of a `CaseEvaluationData`. I think they should be stand-alone methods that accept a `CaseEvaluationData` as the first argument (which is actually what you've written, if you think about it). The logic here is that your `CaseEvaluationData` is just a simple struct to keep related data tagged in one place. The helper methods use `self` to indirectly fetch data. But there's nothing special about the instance of `CaseEvaluationData` that calls them.

So just make these all stand-alone functions, and replace the argument `self` with `case_evaluation_data: CaseEvaluationData`.

src/extremeweatherbench/evaluate.py
```diff
@@ -253,11 +398,15 @@ def _open_obs_datasets(eval_config: config.Config):
...
+    if eval_config.point_obs_path:
+        point_obs = pd.read_parquet(
+            eval_config.point_obs_path, storage_options=POINT_OBS_STORAGE_OPTIONS
+        )
     if eval_config.gridded_obs_path:
         gridded_obs = xr.open_zarr(
             eval_config.gridded_obs_path,
             chunks=None,
-            storage_options=dict(token="anon"),
+            storage_options=GRIDDED_OBS_STORAGE_OPTIONS,
```
darothen, line 409: This circles back to my comment at the top of the file about your global variables. This is a tough call; the "best" way would be to provide a set of default options, and then merge it with user-passed ones. Hard-coded behavior, like the auth mode, will lead to non-obvious failure modes for users down the line, without an easy way to rectify without hacking the library. It _might_ be a good idea to go ahead and provide helper arguments to this method or package within the `config.Config` object the ability for a user to over-ride these storage_options.

If you want to hard-code everything, leave a description in the docstring about this behavior so that at least it's very explicit what is happening.

PR #78, opened 2025-03-12
Modularize data loading

aaTman:
# EWB Pull Request

## Description

This PR moves the bulk of the data loading logic over to data_loader.py. Associated with that are cleanups in evaluate and utils, including a new function that is necessary to convert timedelta64 values to int.

## Type of change

## How Has This Been Tested?

Pytest + pre-commit locally + tested on a jupyter notebook

## Checklist:


src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Args:
+        eval_config: The evaluation configuration.
+        forecast_schema_config: The forecast schema configuration.
+
+    Returns:
+        The opened forecast dataset.
+    """
+    logger.debug("Opening forecast dataset")
+    if "://" in eval_config.forecast_dir:
```
darothen, line 27: You probably need a note in the docstring providing information on how you infer the filesystem/location of the specified data., as well as the expected file formats (from below)

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,161 @@
...
+        ):
+            raise TypeError("Multiple file types found in forecast path.")
+        if "parq" in eval_config.forecast_dir:
+            # kerchunked parq refs seem to consistently need two "compute" calls
+            # the first one doesn't actually load to memory
+            forecast_ds = open_kerchunk_reference(
+                eval_config.forecast_dir, forecast_schema_config
+            )
+        elif "zarr" in eval_config.forecast_dir:
+            forecast_ds = xr.open_zarr(eval_config.forecast_dir, chunks="auto")
```
darothen, line 48: Is the idea here that this will automatically concatenate all of the files found at that directory? I'm just not following the logic with how you differentiate between how you handle different number of file types and what implication that has for the calls you make to read the datasets.

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,161 @@
...
+            raise TypeError(
+                "Unknown file type found in forecast path, only json, parquet, and zarr are supported."
+            )
+    else:
+        raise FileNotFoundError("No files found in forecast path.")
+
+    mapping = {
+        getattr(forecast_schema_config, field.name): field.name
+        for field in dataclasses.fields(forecast_schema_config)
+    }
```
darothen, line 64: Could use a docstring to make it clear that you expect there to be a particular format of data in the forecast_schema_config.

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,161 @@
...
+    return forecast_ds
+
+
+def open_kerchunk_reference(
+    file,
+    forecast_schema_config: config.ForecastSchemaConfig,
+    remote_protocol: str = "s3",
+) -> xr.Dataset:
+    """Open a dataset from a kerchunked reference file in parquet or json format.
+    This has been built for the CIRA MLWP S3 bucket's data, but can work with other
```
darothen, line 85:
```suggestion

    This has been built for the CIRA MLWP S3 bucket's data, but can work with other
```

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,161 @@
...
+    data as well if it is 6 hourly.
+
```
darothen, line 87: Why does it matter if the data is 6-hourly or not? In your explanation of the assumptions here, you should specify what you are expecting, and then you can note the CIRA MLWP bucket (link to the AWS Public Datasets page for it (https://registry.opendata.aws/aiwp/)) satisfies this convention.

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,161 @@
...
+                    "remote_options": {"anon": True},
+                },
+            },
+        )
+    else:
+        raise TypeError(
+            "Unknown file type found in forecast path, only json and parquet are supported."
+        )
+    kerchunk_ds = kerchunk_ds.rename({"time": "lead_time"})
+    kerchunk_ds["lead_time"] = range(0, 241, 6)
```
darothen, line 128: I think it's a bad idea to hard-code this. This should probably stated in the forecast schema config (number of output timesteps and the time interval between them, demanding that it be fixed).

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,161 @@
...
+            eval_config.point_obs_path, storage_options=dict(token="anon")
+        )
+    if eval_config.gridded_obs_path:
+        gridded_obs = xr.open_zarr(
+            eval_config.gridded_obs_path,
+            chunks=None,
+            storage_options=dict(token="anon"),
+        )
+    if point_obs is None and gridded_obs is None:
+        raise FileNotFoundError("No gridded or point observation data provided.")
```
darothen, line 160: Can this statement ever be reached? If either path is invalid, then you will raise an Exception in the relevant if blocks above. Otherwise, you'll always get a data structure back. This goes to our discussion about avoiding return types `Optional[T]` whenever possible, it's simply not the "Pythonic" way to write interfaces. 

I would wrap each attempt to read the data in a try/catch block and re-raise an Exception if there is an issue reading either one.

src/extremeweatherbench/evaluate.py
```diff
@@ -334,6 +329,8 @@ def evaluate(
                 if x
             ]
         ),
+        # TODO(aaTman): #63 loop does not work properly, it does not skip None's outputting the sum of all cases.
+        # Needs to skip None values
```
darothen, line 333: Any reason you need a loop? The summation here is fine, actually more readable than a list comprehension You can re-write it more compactly and readably as the suggestion:

```
sum([
    len(results)
    for results in all_results.values()
    if results is not None
])

File: src/extremeweatherbench/utils.py [hunk 9]
```diff
@@ -558,3 +558,20 @@ def derive_indices_from_init_time_and_lead_time(
...
+        dataset: The input xarray Dataset that uses the schema's variable names.
+
+    Returns:
+        An xarray Dataset with adjusted types.
+    """
+
+    var = dataset["lead_time"]
+    if var.dtype == np.dtype("timedelta64[ns]"):
+        dataset["lead_time"] = (var / np.timedelta64(1, "h")).astype(int)
+    return dataset
```
darothen, line 577: Log a warning or debug statement if you fail to do the conversion due to failing the type check.

A good idiom is to prefix this method's name with "maybe", e.g. "maybe_convert_dataset_to_lead_time". This indicates to the user that it will perform an in-place modification to the dataset its passed, but if it fails some sort of check, it will simply pass back the unaltered original dataset.

darothen (changes requested):
Some quick tweaks to re-review before merging. The mocked tests are super helpful.

src/extremeweatherbench/data_loader.py
aaTman: Yeah this was legacy code I just moved in here which was working, but I cleaned it up significantly so your issue here (which I agree with) should be resolved.

src/extremeweatherbench/data_loader.py
aaTman: Moved it to its own function for readability.

src/extremeweatherbench/data_loader.py
aaTman: Good point, fixing now. I think I see a clear and simple path for this with some existing defaults.

aaTman: This will break if not using CIRA kerchunk. I added an exception to handle this.

src/extremeweatherbench/data_loader.py
aaTman: Agreed. I think this can be removed. I have a solid idea of the outcome (using empty dataframes and dataarrays in place of None's is what I am thinking), but I attempted to refactor the Optionals up and downstream which became a significant endeavor that will require some time to reconcile. It's top-of-mind to code out of EWB and will throw an issue up to capture it.

src/extremeweatherbench/evaluate.py
aaTman: Fixed in pandas refactor #82

src/extremeweatherbench/config.py
```diff
@@ -55,6 +69,11 @@ class Config:
     cache_dir: Optional[Path] = DEFAULT_CACHE_DIR
     gridded_obs_path: str = ARCO_ERA5_FULL_URI
     point_obs_path: str = ISD_POINT_OBS_URI
+    remote_protocol: str = "s3"
+    forecast_source: Optional[str] = "cira"
```
darothen, line 73: Is there any particular reason we need something special here for the CIRA archive? It's more in our interest to force ourselves to use the standard machinery that everyone will use for any forecast data source than to special/hard-code anything here.

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,169 @@
+import xarray as xr
+from extremeweatherbench import config, utils
+import pandas as pd
+import logging
+import dataclasses
+from typing import Tuple, Optional
```
darothen, line 6: nit: import order; be sure that `isort` is installed and enabled in your dev environment. It will group imports by groups and then sort alphabetically. Makes it easier to find things.

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def open_kerchunk_reference(
+    eval_config: config.Config,
+) -> xr.Dataset:
+    """Open a dataset from a kerchunked reference file in parquet or json format.
+    This has been built for the CIRA MLWP S3 bucket's data (https://registry.opendata.aws/aiwp/),
+    but can work with other data in the future. Currently only supports CIRA data unless
+    schema is identical to the CIRA schema.
+
```
darothen, line 54: Keep this function as generic as possible. You can write a helper function, `open_cira_archive`, which _uses_ this function by creating and appropriate `eval_config`. But we should not be hard-coding anything about specific datasets - instead, force ourselves to use the appropriate convention and interface that other users have to.

src/extremeweatherbench/data_loader.py
```diff
@@ -0,0 +1,169 @@
...
+                    "remote_options": {"anon": True},
+                },
+            },
+        )
+    else:
+        raise TypeError(
+            "Unknown kerchunk file type found in forecast path, only json and parquet are supported."
+        )
+    if eval_config.forecast_source == "cira":
+        kerchunk_ds = kerchunk_ds.rename({"time": "lead_time"})
```
darothen, line 94: See prior comments about hard-coding CIRA info in the config. It's okay to have some helper functions on the side, but this should not be in the public interfaces.

src/extremeweatherbench/utils.py
```diff
@@ -558,3 +558,24 @@ def derive_indices_from_init_time_and_lead_time(
     init_time_subset_indices = valid_time_indices[0]
 
     return init_time_subset_indices
+
+
+def maybe_convert_dataset_lead_time_to_int(dataset: xr.Dataset) -> xr.Dataset:
+    """Convert types of variables in an xarray Dataset based on the schema,
+    ensuring that, for example, the variable representing lead_time is of type int.
```
darothen, line 565: nit: Prefix method name with "_" (`_maybe_convert_dataset_lead_time_to_int`) to emphasize it's a private method and users should not attempt to use it unless they know what they're doing.

src/extremeweatherbench/config.py
aaTman: No, a generic preprocessing method that can modify incoming metadata can be incredibly useful here though.

aaTman:
Made some key changes to avoid the hardcoding:

1. We created a `_default_preprocess()` function that acts as a default passthrough. I moved it and the associated type checking to `config.py` as a user would include the function in their config to pass into the evaluation.

2. I moved the `_maybe_convert_dataset_lead_time_to_int()` function out of `utils.py` and into `data_loader.py` as it'll realistically only be used there. I also included the `lead_time` generation based on the config file metadata in here in case a user has something other than ints or timedelta64's (or a `lead_time` coordinate at all) in their dataset. Caveat emptor here, it'll break downstream if lead time isn't already an index coordinate. Could use a better check but not breaking.

3. Moved the tests as needed and renamed functions that needed to be renamed in `evaluate.py`.

darothen (approved):
LGTM!

PR #92, opened 2025-05-15
Consolidate schemas into main config

aaTman:
To simplify metadata flow, the forecast and point observation schema configs are placed in the primary configuration dataclass.


src/extremeweatherbench/config.py
```diff
@@ -2,7 +2,7 @@
 
 import dataclasses
 from pathlib import Path
-from typing import Callable, List, Optional
+from typing import Callable, List, Optional, Union
```
darothen, line 5: In Python >= 3.10, you can just use the "|" operator. So `Union[str, Path]` is equivalent to `str | Path` (or most commonly, you would see `str | None`).

src/extremeweatherbench/config.py
```diff
@@ -186,3 +128,67 @@ def mapped_metadata_vars(self):
...
+        gridded_obs_storage_options: A dictionary of cloud storage options for the gridded
+            observation dataset.
+        point_obs_storage_options: A dictionary of cloud storage options for the point
+            observation dataset.
+    """
+
+    event_types: List[events.EventContainer]
+    output_dir: Union[str, Path] = DEFAULT_OUTPUT_DIR
+    forecast_dir: Union[str, Path] = DEFAULT_FORECAST_DIR
+    cache_dir: Union[str, Path] = DEFAULT_CACHE_DIR
```
darothen, line 175: These should just be `Path`, and you should consider a custom constructor which eagerly casts `str` to this type - possibly also checking that the paths exist and are actually directories.

src/extremeweatherbench/config.py
```diff
@@ -186,3 +128,67 @@ def mapped_metadata_vars(self):
...
+    gridded_obs_path: Union[str, Path] = ARCO_ERA5_FULL_URI
+    point_obs_path: Union[str, Path] = DEFAULT_POINT_OBS_URI
```
darothen, line 177: Because these could be remote URI's starting with something like `s3://` or `gs://`, I think the type should just be `str`. Elsewhere, you can check if they're remote URIs or not and then eagerly cast to `Path`.

src/extremeweatherbench/config.py
```diff
@@ -186,3 +128,67 @@ def mapped_metadata_vars(self):
...
+    remote_protocol: str = "s3"
+    forecast_preprocess: Callable[[xr.Dataset], xr.Dataset] = _default_preprocess
+    init_forecast_hour: int = 0  # The first forecast hour (e.g. Graphcast starts at 6).
```
darothen, line 180: Move trailing comment into class docstring

src/extremeweatherbench/config.py
```diff
@@ -100,7 +42,7 @@ class ForecastSchemaConfig:
     suggested based on the CF Conventions.
     """
 
-    surface_air_temperature: Optional[str] = "t2m"
+    surface_air_temperature: Optional[str] = "t2"
```
darothen, line 45: Why change from "t2m"? The GRIB shortName for 2m air temperature is actually "2t", but I don't believe xarray field names can start with numbers so by convention `cfgrib` reads this as "t2m". I think it's better just to keep "t2m."

darothen (approved):
A few tweaks to consider. but overall good to go.

src/extremeweatherbench/config.py
aaTman: These defaults are meant to follow the CIRA MLWP defaults; I ended up setting this as `t2m` in a prior commit that I've been addressing each time I used the CIRA data during the initialization of the schema. I otherwise agree (and it's why I changed in the past).

PR #94, opened 2025-05-29
Taylor/ewb-19-create-derived-variable-methodology

aaTman:
# EWB Pull Request

## Description

Adding new derived ABC for use in upcoming metrics and event types.

## Type of change

## How Has This Been Tested?

Added pytest unit testing; type checking can change once the ABC is applied more concretely. 

## Checklist:


src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self):
+        pass
+
+    @property
+    def name(self) -> str:
+        """Return the class name without parentheses."""
+        return self.__class__.__name__
+
+    @abc.abstractmethod
+    def calculate(self, data: xr.Dataset | xr.DataArray) -> xr.DataArray:
```
darothen, line 41: I think this is too flexible of a signature; I'd propose that we restrict this to an `xr.Dataset`, and require that all the fields necessary to compute the derived variable must be contained in that structure.

Another thing to consider - we likely need at least a `**kwargs` in the abstract signature, since some methods for deriving variables may have specific parameters that need customization.

darothen (changes requested):
Suggesting a small tweak to simplify signature on your `calculate` interface.

PR #99, opened 2025-07-23
Bounding Box + Shapefile Functionality

aaTman:
# EWB Pull Request

## Description

This PR adds functionality for incoming cases' location attribute.

Now, on top of using a coordinate with defined distances in degrees from the center (as it is currently set up), users can:

1. Input lat and lon corners and

2. Use a shapefile link in their yaml or dictionary calls.

There is a semi-breaking change in that the yaml and dictionary structure must have all location attributes under location, instead of having bounding_box_degrees as a distinct key.

Some of these changes are temporary while the refactor is continuing.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

New tests have been created and old tests continue to work as expected with existing code.


src/extremeweatherbench/case.py
aaTman, line 1: I wouldn't read into this file for long as much of this is going to be dropped in the refactor, such as the redundant code in the heat and freeze case classes.

src/extremeweatherbench/utils.py
```diff
@@ -437,10 +389,10 @@ def align_point_obs_from_gridded(
...
     case_subset_point_obs_df = location_subset_point_obs(
         case_subset_point_obs_df,
-        forecast_ds["latitude"].min().values,
-        forecast_ds["latitude"].max().values,
-        forecast_ds["longitude"].min().values,
-        forecast_ds["longitude"].max().values,
+        float(forecast_ds["latitude"].min().values),
+        float(forecast_ds["latitude"].max().values),
+        float(forecast_ds["longitude"].min().values),
+        float(forecast_ds["longitude"].max().values),
```
aaTman, line 395: This block will be removed for the new observation refactor; only edited to prevent tests from failing

src/extremeweatherbench/evaluate_cli.py
```diff
@@ -60,14 +61,19 @@ def event_type_constructor(loader: yaml.SafeLoader, node: yaml.nodes.MappingNode
...
 
-        cases = [
-            case.IndividualCase(**yaml_event_case["cases"][i - 1]) for i in case_ids
-        ]
+        cases = []
+        for i in case_ids:
+            case_data = yaml_event_case["cases"][i - 1].copy()
+            # Convert location dict to Region object
+            case_data["location"] = regions.map_to_create_region(case_data["location"])
+            cases.append(case.IndividualCase(**case_data))
```
aaTman, line 69: Will be reviewing this approach with the refactor, very likely this will change.

src/extremeweatherbench/utils.py
```diff
@@ -44,56 +42,20 @@ def convert_longitude_to_360(longitude: float) -> float:
...
+) -> float | Union[xr.Dataset, xr.DataArray]:
+    """Convert a longitude from the range [0, 360) to [-180, 180).
+
+    Datasets are coerced to [-180, 180) and sorted by longitude."""
+    if isinstance(longitude, xr.Dataset) or isinstance(longitude, xr.DataArray):
+        longitude.coords[longitude_name] = (
+            longitude.coords[longitude_name] + 180
+        ) % 360 - 180
+        longitude = longitude.sortby(longitude_name)
+        return longitude
```
aaTman, line 56: the removed code here was shifted to `regions.py`. The addition here allows for `convert_longitude_to_180` to handle a float OR the xarray case, where the longitude then need to be sorted to work properly for downstream operations.

src/extremeweatherbench/regions.py
```diff
@@ -0,0 +1,240 @@
...
+def create_region(
+    latitude: Optional[float] = None,
+    longitude: Optional[float] = None,
+    bounding_box_degrees: Optional[Union[float, Tuple[float, float]]] = None,
+    latitude_min: Optional[float] = None,
+    latitude_max: Optional[float] = None,
+    longitude_min: Optional[float] = None,
+    longitude_max: Optional[float] = None,
+    shapefile_path: Optional[Union[str, Path]] = None,
+) -> Region:
```
aaTman, line 145: Couldn't think of a nicer way to manage this factory function, but open to thoughts now or in the future!

src/extremeweatherbench/regions.py
```diff
@@ -0,0 +1,240 @@
...
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+class Region:
+    """Base class for different region representations."""
+
+    def to_geopandas(self):
+        """Convert the region to a geopandas object."""
+        if isinstance(self, ShapefileRegion):
```
darothen, line 25: Seeing how this is used later, why not just create a `geopandas` property? E.g.,

``` python
class Region:

    @property
    def geopandas(self):
        # construct GeoDataFrame here
        return gdf
```

Technically this would be an abstract property and each implementation would provide the logic.

src/extremeweatherbench/regions.py
```diff
@@ -0,0 +1,240 @@
...
+        else:
+            raise NotImplementedError(f"to_geopandas not implemented for {type(self)}")
+
+    def mask(self, dataset: xr.Dataset) -> xr.Dataset:
+        """Mask a dataset to the region."""
+        mask = regionmask.mask_geopandas(
+            self.to_geopandas(), dataset.longitude, dataset.latitude
+        )
+        return dataset.where(mask)
+
```
darothen: Following our chat this morning, I think this should be an ABC with some stubs that each implementation should handle:


```suggestion
from abc import ABC, abstractmethod

class Region(ABC):
    """Base class for different region representations."""
    
    @classmethod
    @abstractmethod
    def create_region(cls, *args, **kwargs):
        """Abstract factor method to create a region; subclasses must implement with their own, specialized arguments."""
        pass
    
    @property
    @abstractmethod
    def geopandas(self) -> gpd.GeoDataFrame:
        """Return representation of this Region as a GeoDataFrame"""
        pass

    # NOTE: now everyone gets this method for free!
    def mask(self, dataset: xr.Dataset) -> xr.Dataset:
        """Mask a dataset to the region."""
        mask = regionmask.mask_geopandas(
            self.geopandas(), dataset.longitude, dataset.latitude
        )
        return dataset.where(mask)

```

Note that to implement the interface, a subtype Region can implement its factory constructor arguments however it wants:

``` Python
class CenteredRegion(Region):
    
    def __init__(self, lat: float, lon: float, bbox_degrees: Union[float, Tuple[float, float]]):
        ....

    @classmethod
    def create_region(cls, lat: float, lon: float, bbox_degrees: Union[float, Tuple[float, float]]):
        return cls(lat, lon, bbox_degrees)
```

src/extremeweatherbench/regions.py
```diff
@@ -0,0 +1,240 @@
...
+    elif all(
+        x is not None
+        for x in [latitude_min, latitude_max, longitude_min, longitude_max]
+    ):
+        # Type checkers can't infer that the values are not None after the all() check
+        # so we need to assert or cast them
+        assert latitude_min is not None
+        assert latitude_max is not None
+        assert longitude_min is not None
+        assert longitude_max is not None
```
darothen, line 187: Simpler ->
```suggestion
        if None in [latitude_min, latitude_max, longitude_min, longitude_max]:
            raise ValueError(...)
```

src/extremeweatherbench/regions.py
```diff
@@ -0,0 +1,240 @@
...
+        raise ValueError(
+            "Invalid parameters. Must provide either (latitude, longitude, bounding_box_degrees) "
+            "or (latitude_min, latitude_max, longitude_min, longitude_max) or shapefile_path."
+        )
+
+
+def clip_dataset_to_bounding_box_degrees(
+    dataset: xr.Dataset, location: CenteredRegion
+) -> xr.Dataset:
+    """Clip an xarray dataset to a box around a given location in degrees latitude & longitude.
```
darothen, line 204: If `location` must be a **CenteredRegion**, then why not just make this an instance method of **CenteredRegion**? Also, any reason that this can't just be your **Region.mask()** function followed by an additional dropna on the dataset?

src/extremeweatherbench/case.py
```diff
@@ -40,18 +38,62 @@ class IndividualCase:
...
+                    ),
+                }
+            )
+        elif isinstance(self.location, regions.ShapefileRegion):
+            region_mask = regionmask.from_geopandas(self.location.shapefile)
+            modified_ds = region_mask.mask(
+                dataset.rename({longitude_name: "lon", latitude_name: "lat"})
+            )
+        else:
+            raise ValueError(f"Unsupported location type: {type(self.location)}")
```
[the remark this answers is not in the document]
aaTman: this is deprecated code; simplified drastically with new feedback

src/extremeweatherbench/regions.py
aaTman: Done! This was indirectly an artifact of trying to subset the ARCO ERA5 when chunks=None, where the data was loaded into memory.

src/extremeweatherbench/regions.py
```diff
@@ -0,0 +1,301 @@
...
+
+    Args:
+        region_input: Either a Region object or a dictionary of parameters.
+
+    Returns:
+        A Region object.
+    """
+    if isinstance(region_input, Region):
+        return region_input
+    elif isinstance(region_input, dict):
```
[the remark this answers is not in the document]
aaTman: good catch - artifact of an else raise valueerror that wasn't possible to get to in the first place.

src/extremeweatherbench/regions.py
```diff
@@ -299,3 +262,11 @@ def _check_longitude_bounds(
...
         )
     return longitude_min, longitude_max
+
+
+# Registry of region types that can be extended by users
+REGION_TYPES: dict[str, Type[Region]] = {
+    "centered_region": CenteredRegion,
+    "bounded_region": BoundingBoxRegion,
+    "shapefile_region": ShapefileRegion,
+}
```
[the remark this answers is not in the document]
aaTman: I don't know if this is possible as the classes weren't defined yet. Is there an alternative?

PR #108, opened 2025-08-11
EWB Refactor

aaTman:
# EWB Pull Request

## Description

The philosophy behind how the EWB code is set up is modularity, user friendliness, and as transparent compute requirements as possible. Simply put: _A user should be able to build an evaluation with ease that scales relative to the size+number of the cases they want to evaluate._

ExtremeWeatherBench’s library is built such that a user provides a configuration to a class ExtremeWeatherBench that drives the evaluation of a set of cases. The entities that drive ExtremeWeatherBench are **Forecasts, Targets, Cases, Variables/DerivedVariables, and Metrics**. A very high level overview of the code flow:

**Case Metadata** includes the subsetting information provided within a yaml file or dictionary that includes ID number, name, event type, start date, end date, and location. 

**MetricEvaluations** include a list of one or more metrics, a target configuration, forecast configuration, and event type. It is a configurable dataclass that can be stored or built on the fly. Target and forecast configurations define the dataset source, access options, and what variables/DerivedVariables will be used in the evaluation.

The combination of cases and events are called CaseOperators. CaseOperators are distinct dataclasses which contain all of the information needed to produce an evaluation:

- Location
- Time
- Event Type
- Metric(s)
- A Forecast 
- A Target
- Variables and/or Derived Variables

A user can instantiate an ExtremeWeatherBench class with the case metadata and MetricEvaluations and then use the run() method to execute the code, resulting in a pandas DataFrame output of the metrics and the associated metadata. The internal loop operates as follows:

> For each case operator 
> &nbsp;&nbsp;&nbsp;&nbsp;Build the target and forecast xarray datasets, including derived variables if in evaluation*
> &nbsp;&nbsp;&nbsp;&nbsp;Align them
>&nbsp;&nbsp;&nbsp;&nbsp;For the product of each zipped variable combination and metric
>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Evaluate the metric
> &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Append a DataFrame with results

* Building the target and forecast datasets share a run_pipeline function that:
- Opens and possible preprocesses the source data (if a preprocess function is provided)
- Maps variable naming based on provided mapping dictionary
- Spatiotemporally subsets the dataset to the case metadata
- Converts to an xarray Dataset if not already
- Derives variables if any in variables
- Adds the source name to the Dataset attributes for use in the output DataFrame

Each step is provided by required methods in each Target and Forecast class. Derived variables also require a similar approach to calculation.

Users can also opt to parallelize the CaseOperators but is risky (until I figure out a fix or recommended approach) due to the greedy nature of dask.load/dask.compute with xarray and/or zarr.

## Checklist Before Ready for Merge for Baseline Pre-Release EWB:
- Command Line Interface Update
- Thorough Testing

### DIFF
--- a/tests/test_evaluate_cli.py
+++ b/tests/test_evaluate_cli.py
@@ -258,7 +258,8 @@ class TestCaseOperatorSaving:
         assert result.exit_code == 0
         # Output suppressed - only check exit code
 
-        # Verify pickle file was created and contains the right data (in temp dir, auto-cleanup)
+        # Verify pickle file was created and contains the right data (in temp dir,
+        # auto-cleanup)
         assert save_path.exists()
         with open(save_path, "rb") as f:
             loaded_ops = pickle.load(f)
@@ -347,7 +348,8 @@ class TestValidationAndErrorHandling:
         with runner.isolated_filesystem():
             result = runner.invoke(evaluate_cli.cli_runner, ["--default"])
             assert result.exit_code == 0
-            # Check that the CLI completed successfully - any files created are in the isolated temp filesystem
+            # Check that the CLI completed successfully - any files created are in the
+            # isolated temp filesystem
 
 
 class TestResultsSaving:


src/extremeweatherbench/case.py
```diff
@@ -30,165 +32,99 @@ class IndividualCase:
...
+    Args:
+        cases_dict: The case metadata to use for the case operators.
+        metric_evaluation_objects: The metric evaluation objects to use for the case operators.
 
-    HEAT_WAVE = "heat_wave"
-    FREEZE = "freeze"
+    Returns:
+        A list of CaseOperator objects.
+    """
+    case_metadata_collection = dacite.from_dict(
```
gideonite, line 87: Sorry for being vague but this just doesn't feel right. Generally, parsing of config(s) should happen once, at the top-level entry point to the system.

src/extremeweatherbench/defaults.py
```diff
@@ -0,0 +1,236 @@
...
+era5_heatwave_target = inputs.ERA5(
+    source=inputs.ARCO_ERA5_FULL_URI,
+    variables=["surface_air_temperature"],
+    variable_mapping={
+        "2m_temperature": "surface_air_temperature",
+        "time": "valid_time",
+    },
+    storage_options={"remote_options": {"anon": True}},
+)
+
```
gideonite, line 13: nit: if it were me, I would remove white space here to group these together.

src/extremeweatherbench/defaults.py
```diff
@@ -0,0 +1,236 @@
...
+        "surface_temperature": "surface_air_temperature",
+        "surface_eastward_wind": "surface_eastward_wind",
+        "surface_northward_wind": "surface_northward_wind",
+    },
+    storage_options={"remote_options": {"anon": True}},
+)
+
+# LSR/PPH target
+
+# TODO: add PPH target
```
gideonite, line 72: Nit: better to pull this TODO into an issue or remove it entirely. You can grep your codebase for TODOs: `grep -i todo`. Or, use ripgrep, a tool I much prefer: https://github.com/BurntSushi/ripgrep

src/extremeweatherbench/defaults.py
```diff
@@ -0,0 +1,236 @@
...
+lsr_target = inputs.LSR(
+    source=inputs.LSR_URI,
+    variables=["local_storm_reports"],
+    variable_mapping={},
+    storage_options={"remote_options": {"anon": True}},
+)
+
+# IBTrACS target
```
gideonite, line 80: I'm an outsider. I have no idea what this means, but I assume this has to do with TC tracks. If it were possible to make this a general descriptive name, I would be all for it.

src/extremeweatherbench/defaults.py
```diff
@@ -0,0 +1,236 @@
...
+    variable_mapping={
+        "10u": "surface_eastward_wind",
+        "10v": "surface_northward_wind",
+        "z": "geopotential",
+        "msl": "pressure_at_mean_sea_level",
+    },
+    storage_options={"remote_protocol": "s3", "remote_options": {"anon": True}},
+)
+
+BRIGHTBAND_EVALUATION_OBJECTS = [
```
gideonite, line 161: I'm surprised by the "sudden" global variable here.

src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,205 @@
...
+        Args:
+            data: The dataset to build the derived variable from.
+
+        Returns:
+            A DataArray with the derived variable.
+        """
+        self._check_data_for_variables(data)
+        return self.derive_variable(data)
+
+    def _check_data_for_variables(self, data: xr.Dataset):
```
gideonite, line 72: Can this be public? It could be called `validate` or something like that.

src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,205 @@
...
+        """Check that the data has the variables required to build the variable, based on assigned input variables."""
+        for v in self.input_variables:
+            if v not in data.data_vars:
+                raise ValueError(f"Input variable {v} not found in data")
+
+
+class CravenSignificantSevereParameter(DerivedVariable):
+    """A derived variable that computes the Craven significant severe parameter."""
```
gideonite, line 80: Ideally, this would include a link to wikipedia, arxiv paper, etc. This looks like as good of a link as any: https://www.spc.noaa.gov/exper/mesoanalysis/help/help_cbsig.html

src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,205 @@
...
+        "2m_dewpoint_temperature",
+        "2m_relative_humidity",
+        "10m_u_component_of_wind",
+        "10m_v_component_of_wind",
+        "surface_pressure",
+        "geopotential",
+    ]
+
+    def derive_variable(self, data: xr.Dataset) -> xr.DataArray:
+        """Derive the Craven significant severe parameter."""
```
gideonite, line 94: nit: parameter vs. variable. Is it a parameter or a variable?

src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,205 @@
...
+        # cbss_ds = calc.craven_brooks_sig_svr(data[variables],variable_mapping={'pressure':'level', 'dewpoint':
+        # 'dewpoint_temperature','temperature':'air_temperature'})
```
gideonite, line 96: Generally, commented out code is not great. Can it be removed without causing a great deal of pain?

src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,205 @@
...
+        return test_da
+
+
+class SurfaceWindSpeed(DerivedVariable):
+    """A derived variable that computes the surface wind speed."""
+
+    name = "surface_wind_speed"
+    input_variables = ["10u", "10v"]
+
+    def derive_variable(self, data: xr.Dataset) -> xr.DataArray:
```
gideonite, line 107: Aha, so these are essentially wrappers of derived variable computations. They are pure mathematical functions. 

I would consider using python's `__call__` interface.

src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,205 @@
...
+        """Derive the surface wind speed."""
+        return np.sqrt(
+            data[self.input_variables[0]] ** 2 + data[self.input_variables[1]] ** 2
+        )
```
gideonite, line 111: Are the units of the results array preserved?

src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,205 @@
...
+
+class AtmosphericRiverMask(DerivedVariable):
+    """A derived variable that computes the atmospheric river mask."""
+
+    name = "atmospheric_river_mask"
+    input_variables = ["msl"]
+
+    def derive_variable(self, data: xr.Dataset) -> xr.DataArray:
+        """Derive the atmospheric river mask."""
+        return data[self.input_variables[0]] < 1000
```
gideonite, line 122: I don't totally understand why this is better than just doing:
```
data["msl"] < 1000
```

src/extremeweatherbench/derived.py
```diff
@@ -0,0 +1,205 @@
...
+            data[self.input_variables[0]]
+            * data[self.input_variables[1]]
+            * data[self.input_variables[2]]
+        )
+
+
+class IntegratedVaporTransportJacobian(DerivedVariable):
+    """A derived variable that computes the integrated vapor transport Jacobian."""
+
+    name = "integrated_vapor_transport_jacobian"
```
gideonite, line 143: This is a jacobian? Wat? Sorry, I'm just confused, has nothing to do with the code.

src/extremeweatherbench/utils.py
```diff
@@ -3,22 +3,27 @@
 """
```
gideonite, line 3: I'm fighting a loosing battle but I'll make an attempt anyway: 

"Just say no to utils files."

Why? They contain a bunch of unrelated code that is used randomly throughout the codebase -- basically a junk drawer. New contributors won't know to look here for some bit of functionality. ChatGPT says: "hides intent, grows without bounds, increases merge conflicts, and makes refactoring harder." Better to organize code by what it does, rather than what it does *not* do, e.g. doesn't fit elsewhere.

Usually, functions here will only be used in one or two places. I haven't checked yours but worth giving it a thought. If indeed this is the case, then better to just "in-line" those functions closer to the call site. If you need the function in two places, then just import it once. At some point, it breaks down and it makes sense to put it in a third place at which point it is hopefully clear what that place is --- "data_helpers.py," "math.py," etc.

src/extremeweatherbench/utils.py
```diff
@@ -578,10 +579,204 @@ def maybe_convert_to_path(value: str | Path) -> str | Path:
...
+    This method uses introspection to determine which arguments the callable
+    can accept and filters kwargs accordingly.
+
+    Args:
+        kwargs: The full kwargs dictionary to filter
+        callable_obj: The callable (function, method, etc.) to check against
+
+    Returns:
+        A filtered dictionary containing only the kwargs that the callable can accept
+    """
```
gideonite, line 599: A private method with a doc string is a bit of a code smell. Should probably be public.

src/extremeweatherbench/utils.py
```diff
@@ -578,10 +579,204 @@ def maybe_convert_to_path(value: str | Path) -> str | Path:
...
+
+
+def maybe_map_variable_names(
+    data: IncomingDataInput, variable_mapping: Optional[dict] = None
+) -> IncomingDataInput:
+    """Map the variable names to the target data, if required.
+
+    Args:
+        data: The incoming data in the form of an object that has a rename method for data variables/columns.
+        variable_mapping: The mapping of variable names to the incoming data, with the format {incoming_name: new_name}.
```
gideonite, line 690: Super long lines. Is there a linter on this codebase? Will be important as contributions accelerate.

src/extremeweatherbench/inputs.py
```diff
@@ -0,0 +1,643 @@
...
+    ) -> utils.IncomingDataInput:
+        if not isinstance(target_data, pd.DataFrame):
+            raise ValueError(f"Expected pandas DataFrame, got {type(target_data)}")
+
+        # latitude, longitude are strings by default, convert to float
+        target_data["latitude"] = target_data["latitude"].astype(float)
+        target_data["longitude"] = target_data["longitude"].astype(float)
+        target_data["valid_time"] = pd.to_datetime(target_data["valid_time"])
+
+        filters = (
```
gideonite, line 387: This large block could use a comment.

src/extremeweatherbench/derived.py
aaTman: I think one of the functions (yet to include) has a link to the paper as well where this originates from.

src/extremeweatherbench/derived.py
aaTman: This is going to be uncommented; currently in-flight to get this connected in.

src/extremeweatherbench/derived.py
aaTman: These all are TODO's, let me update this now.

src/extremeweatherbench/derived.py
aaTman: it will be; sorry - in flight to add this!

src/extremeweatherbench/derived.py
aaTman: oh interesting. let me look into this and see if it's an easy swap.

src/extremeweatherbench/defaults.py
aaTman: this is meant to be the reference configuration list to run our builtin suite of eval metrics with. I'm using this file as a way for users to easily access these recipes, more or less.

src/extremeweatherbench/calc.py
```diff
@@ -0,0 +1,504 @@
...
+        ) / 9.80665
+        geopotential_thickness.attrs = dict(
+            description="Geopotential thickness of level and 500 hPa", units="m"
+        )
+        return geopotential_thickness
+
+    def generate_variables(self, ds: xr.Dataset):
+        """Subset the variables based on the variables argument.
+
+        Using "min" will subset the minimum required variables - sea level pressure and geopotential thickness.
```
hansmohrmann, line 91: What are these in reference to? Maybe a deleted function argument?

src/extremeweatherbench/calc.py
```diff
@@ -0,0 +1,504 @@
...
+import pandas as pd
+import xarray as xr
+from scipy import spatial
+from skimage import measure
+from skimage.feature import peak_local_max
+
+from extremeweatherbench import inputs
+
+
+class CycloneDatasetBuilder:
```
hansmohrmann, line 17: suggestion: as far as I can tell, this class just contains 4 methods that are used exclusively within the derived.TCTrack class. Any reason you couldn't either spin these methods into indepedent functions, or else incorporate them into derived.TCTrack

src/extremeweatherbench/calc.py
```diff
@@ -0,0 +1,504 @@
...
+        ds_mapping.isel(
+            latitude=int(input_point[0]), longitude=int(input_point[1])
+        ).latitude.values,
+        ds_mapping.isel(
+            latitude=int(input_point[0]), longitude=int(input_point[1])
+        ).longitude.values,
+    )
+
+
+def calculate_haversine_degree_distance(
```
hansmohrmann, line 205: Would trust with amply unit tests!

src/extremeweatherbench/calc.py
```diff
@@ -0,0 +1,504 @@
...
+            track_id += 1
+    return tracks
+
+
+def create_tctracks_from_dataset(
+    cyclone_dataset: xr.Dataset,
+    slp_contour_magnitude=200,
+    dz_contour_magnitude=-6,
+    min_distance=12,
+):
```
hansmohrmann, line 394: return type needed. This looks like a known algorithm for determining a TC track and would benefit from a descriptive docstring.

src/extremeweatherbench/defaults.py
hansmohrmann, line 1: Love having a large exhaustive defaults here! I think a user workflow that doesn't involve more than just configuring the details of their own forecast is definitely a necessity and that relies on having all these defaults preconfigured.

dev/refactor_scripts.py
hansmohrmann, line 1: what is this module for? is it temporary, or part of the library?

src/extremeweatherbench/calc.py
aaTman: Yeah, that's what I'm thinking. This is an in-flight port of the one-off MWE I set up which I'm currently generalizing to the repo.

src/extremeweatherbench/utils.py
aaTman: Yeah; set to 120 characters. Is there a BB standard?

src/extremeweatherbench/calc.py
aaTman: unit tests are on the way soon! fully agree.

src/extremeweatherbench/calc.py
```diff
@@ -0,0 +1,594 @@
...
+from skimage.feature import peak_local_max
+
+from extremeweatherbench import inputs
+
+Location = namedtuple("Location", ["latitude", "longitude"])
+
+
+# TODO: determine if this is needed or can be removed for the refactor
+@dataclasses.dataclass
+class TC:
```
darothen, line 21: nit: prefer more explicit class names, like `TropicalCyclone`. In this particular case, it's really just a struct wrapping a single timestamped observation, so it should be something like `TropicalCycloneObservation`

src/extremeweatherbench/calc.py
```diff
@@ -0,0 +1,594 @@
...
+    id: int
+    valid_time: pd.Timestamp
+    coordinate: Location
+    vmax: float
+    slp: float
+
+
+# TODO: determine if this is needed or can be removed for the refactor
+@dataclasses.dataclass
+class TCTracks:
```
darothen, line 31: This is just a single track though, isn't it? The plural naming is confusing. Is it tracks from multiple cyclones or the track timestamps of a single cyclone, or something else?

src/extremeweatherbench/calc.py
aaTman: 100%. This is inflight from the MWE. Once we have the flow working end to end I will get these functions all cleaned up code and docstring-wise.

src/extremeweatherbench/metrics.py
```diff
@@ -1,180 +1,446 @@
...
+        return {
+            "forecast": subset_forecast,
+            "target": max_min_target_value,
+            "preserve_dims": self.base_metric().preserve_dims,
+        }
+
+
+class OnsetME(AppliedMetric):
+    base_metric = ME
+    preserve_dims: str = "init_time"
```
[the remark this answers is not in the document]
aaTman: Good question. My approach is to consider temporal reduction operations as metrics and many-to-many or single-to-single operations as in scope for a derived variable, I think.

PR #110, opened 2025-08-14
Forecast/Target Alignment

aaTman:
# EWB Pull Request

## Description

This PR sets up a more comprehensive set of alignments for forecasts and target data, which didn't exist prior.


aaTman:
Also expected to fail pre-commit and tests.

PR #111, opened 2025-08-14
Simplify DerivedVariable

aaTman:
# EWB Pull Request

## Description

This PR applies the changes to simplify the derivedvariable workflow and assigning derivedvariables in the workflow after optional computation and after subsetting to ideally improve on the dask graph structure. Tests are also included as well as a new atmospheric river script still in-flight once some load-in-memory issues are handled.


aaTman:
These tests are expected to fail as they're part of the incomplete feature branch.

PR #165, opened 2025-09-27
Flexible case collection region subsetting

aaTman:
# EWB Pull Request

## Description

We've run into the use case multiple times where a user needs to have flexibility in subsetting an existing case collection (e.g. the default we intend to use) to a region such as North America. 

This PR adds functionality in 3 places to do so: at the point where a case is loaded into the script, as an argument in ExtremeWeatherBench, and as a way to subset a completed evaluation's pandas DataFrame. 

The driver for this approach is the `RegionSubsetter` class, which accepts a Region class or lat/lon bounds as the area to subset, as well as a `method` attribute that defines how the cases are subset. This new class hooks into the existing Region framework easily and allows for more flexibility and options for users to run analyses.

## Type of change

## How Has This Been Tested?

New tests have been set up and tested against the new code successfully.

## Checklist:

### DIFF
--- a/src/extremeweatherbench/evaluate.py
+++ b/src/extremeweatherbench/evaluate.py
@@ -61,9 +61,7 @@ class ExtremeWeatherBench:
     # them outside the class if desired for a workflow outside of the class
     @property
     def case_operators(self) -> list["cases.CaseOperator"]:
-        """Case operators as a property are a convenience method for users to use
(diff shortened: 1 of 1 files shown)


darothen (approved):
LGTM

darothen:
@aaTman bumping the LGTM here

PR #170, opened 2025-10-02
Add generate_lsr to data_prep

aaTman:
# EWB Pull Request

## Description

To use severe convective events, we download the SPC filtered local storm reports and generate a practically perfect hindcast using reports for hail and tornado. This PR includes the code used to pull that data.


darothen (approved):
LGTM

PR #180, opened 2025-10-10
Handle alternative/optional variables in derived vars

aaTman:
# EWB Pull Request

## Description

DerivedVariables can have a few different input variables to be successful. Some models do not output the direct variable needed, but instead others that can be used to get to the same outcome (e.g. specific humidity vs relative humidity and temperature). 

Two optional attributes for DerivedVariables are added here and handled - `alternative_variables` and `optional_variables`.

`alternative_variables` is a dictionary that maps the `required_variable` to the variables needed to compute it to get to the same outcome.

`optional_variables` includes variables that, if not included, would not break a derived variable being generated but could slow it down in order to generate or process it, e.g. if a forecast already has orography in its data variables.

Each source file contains logic to catch these mappings.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.6.0
+    rev: v0.14.2
     hooks:
       # Run the linter.
       - id: ruff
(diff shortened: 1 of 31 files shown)


darothen (changes requested):
I think this generally works. It's rather complex... would it not be simpler to just force an invariant that the user must pre-populate the dataset they pass into EWB with all the variables they need for a computation? This could easily be wired into the pre-process apparatus that you already built. The result would be significantly reduced complexity.

aaTman:
> I think this generally works. It's rather complex... would it not be simpler to just force an invariant that the user must pre-populate the dataset they pass into EWB with all the variables they need for a computation? This could easily be wired into the pre-process apparatus that you already built. The result would be significantly reduced complexity.

Ideally users should be responsible for having the variables ready to go ahead of time or using a preprocessor. My philosophy to add this, which I agree is a little more complex than ideal, is we keep running into scenarios with existing models like on the CIRA archive where one variable was available, while the other was not. If there was a community addition of a DerivedVariable, it enables some degree of abstraction instead of requiring each new DerivedVariable to have users create a new preprocessing function.

Let me spend a little time today toying with preprocessing functions to see if it can be used effectively in place of this paradigm.

aaTman:
For brevity, the issue that warrants this approach is preprocessing ARCO ERA5.  

We default to have `chunks=None` (as it is significantly faster this way). If we assign or modify the dataset before variable and spatiotemporal subsetting, it will throw an OOM without chunking beforehand.

All of my experiments with different chunks have resulted in either significantly slower load times or unexplainable stalls in the code, e.g. during computation of the Laplacian in the AR code w/ scipy.ndimage applied to an `xr.apply_ufunc` that worked fine beforehand.

PR #181, opened 2025-10-10
Catch empty targets and/or forecasts earlier in pipeline

aaTman:
# EWB Pull Request

## Description

The current logic in EWB catches zero length dims in the forecast and returns two empty Datasets.

This new method is a bit more robust and will capture more quickly if either there's no valid time or no spatial data available for the case in either target or forecast data.

### DIFF
--- a/tests/test_evaluate.py
+++ b/tests/test_evaluate.py
@@ -1157,7 +1157,7 @@ class TestPipelineFunctions:
             assert isinstance(target_ds, xr.Dataset)
 
             # Should log a warning
-            mock_warning.assert_called_once()
+            mock_warning.assert_called()
             warning_message = mock_warning.call_args[0][0]
             assert "has no data for case time range" in warning_message
             assert (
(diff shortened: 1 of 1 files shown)


src/extremeweatherbench/sources/__init__.py
```diff
@@ -1,13 +1,37 @@
...
     "safely_pull_variables_pandas_dataframe",
+    "check_for_valid_times_xr_dataset",
+    "check_for_valid_times_xr_dataarray",
+    "check_for_valid_times_polars_lazyframe",
+    "check_for_valid_times_pandas_dataframe",
+    "check_for_spatial_data_pandas_dataframe",
+    "check_for_spatial_data_polars_lazyframe",
+    "check_for_spatial_data_xr_dataarray",
+    "check_for_spatial_data_xr_dataset",
 ]
```
darothen: Delete to remove public interface

```suggestion
```

PR #184, opened 2025-10-15
Subset regions more efficiently

aaTman:
# EWB Pull Request

## Description

The current approach to masking regions is simple but has some pitfalls in cases like where the ARCO ERA5 data is not chunked yet, forcing to load the data into memory when the mask is called. 

This PR sets up the code to take advantage of `.sel` in xarray in `.mask`, as well as improves docstrings and style.

Now, a `Region` has a default `mask` operation and `ShapefileRegion` has a custom `mask` implementation to first subset the extrema, then mask to avoid triggering compute as much as possible.

The API doesn't change, but `drop` is no longer used unless in `ShapefileRegion`.

### DIFF
--- a/src/extremeweatherbench/regions.py
+++ b/src/extremeweatherbench/regions.py
@@ -5,11 +5,11 @@ import logging
 import pathlib
 from typing import TYPE_CHECKING, Literal, Mapping, Type, Union
 
-import geopandas as gpd  # type: ignore[import-untyped]
+import geopandas as gpd
 import numpy as np
-import pandas as pd  # type: ignore[import-untyped]
-import regionmask  # type: ignore[import-untyped]
-import shapely  # type: ignore[import-untyped]
+import pandas as pd
+import regionmask
+import shapely
 import xarray as xr
 
 from extremeweatherbench import utils
@@ -46,13 +46,14 @@ class Region(abc.ABC):
             The subset dataset.
         """
         import warnings
+
         ...
 
         if drop:
             warnings.warn(
                 "`drop` is no longer used and will be removed in a future version.",
                 DeprecationWarning,
-                stacklevel=2
+                stacklevel=2,
             )
         longitude_min, latitude_min, longitude_max, latitude_max = (
             self.as_geopandas().total_bounds


src/extremeweatherbench/regions.py
```diff
@@ -1,55 +1,84 @@
...
+        """Get the bounding coordinates of the region.
+
+        Returns:
+            A named tuple with longitude_min, latitude_min, longitude_max, and
+            latitude_max.
+        """
         return namedtuple(
             "BoundingCoordinates",
             ["longitude_min", "latitude_min", "longitude_max", "latitude_max"],
         )(*self.geopandas.total_bounds)
```
darothen: The typing specification is wrong here. If you want to keep things very simple, then the return type could simply be `tuple[int, int, int, int]`, and you would just pass the ints back directly. If you do want to use a struct, then you should do two things:

1. Move the `BoundingCoordinates` definition (the `namedtuple` call) to the module-level
2. Change the return type to `BoundingCoordinates`.

src/extremeweatherbench/regions.py
```diff
@@ -1,55 +1,84 @@
...
+                latitude=slice(latitude_max, latitude_min),
+                longitude=slice(longitude_min, longitude_max),
+            )
+        else:
+            # If monotonically increasing, slice normally (min, max)
+            dataset = dataset.sel(
+                latitude=slice(latitude_min, latitude_max),
+                longitude=slice(longitude_min, longitude_max),
+            )
+
```
darothen: This is a good example of where enforcing a standard on the input data makes life simpler. Pick either convention of "latitudes always increase from -90" or "latitudes always decrease from 90", enforce that at the dataset creation level (you can add sanity checks to warn the user that the data is formatted incorrectly right after the data is passed in to the evaluate function), and then eliminate branching code paths like this. You can still keep a sanity check, but do not give the user more than one way to do things.

src/extremeweatherbench/regions.py
```diff
@@ -173,29 +217,66 @@ class ShapefileRegion(Region):
...
+                latitude=slice(latitude_max, latitude_min),
+                longitude=slice(longitude_min, longitude_max),
+                drop=drop,
+            )
+        # Subset dataset after cutting out a box to minimize memory pressure
+        mask = regionmask.mask_geopandas(
+            self.geopandas, dataset.longitude, dataset.latitude
+        )
+        mask_array = ~np.isnan(mask)
+        return dataset.where(mask_array, drop=drop)
```
darothen: Unless you plan on doing something with `mask_array`, stylistically it's a bit more readable to simply apply it in-line with the return statement.

```suggestion
        return dataset.where(~np.isnan(mask), drop=drop)
```

darothen (changes requested):

PR #206, opened 2025-10-29
Remove `AppliedMetric` and simplify `metrics.py` classes

aaTman:
# EWB Pull Request

## Description

`AppliedMetric` has been a wrapper for `BaseMetric` for a significant portion of EWB's development. In reality it's unnecessary, adding extra code to do something that can be done within a BaseMetric-inherited class. 

All classes inheriting `AppliedMetric`s are now inheriting `BaseMetric`s or the metric they eventually roll up using (e.g. `MaximumMAE` inherits from `MAE`).

> In another PR, `ThresholdMetric` is a subclass of `BaseMetric` that has unique extra requirements.

Also, this PR shifts away from the `classmethod` paradigm and instead simply handles instantiation in the evaluation, if it isn't instantiated in a user-provided script. 

Finally, metric `name` is now set to revert to the class name if not explicitly set as an attribute. It allows a lever of control for users but otherwise is abstracted away.

## Type of change

### DIFF
--- a/src/extremeweatherbench/metrics.py
+++ b/src/extremeweatherbench/metrics.py
@@ -140,7 +140,7 @@ class EarlySignal(BaseMetric):
         raise NotImplementedError("EarlySignal is not implemented yet")
 
 
-class MaximumMAE(BaseMetric):
+class MaximumMAE(MAE):
     """MAE of the maximum value in a tolerance window.
 
     Computes the MAE between the forecast and target maximum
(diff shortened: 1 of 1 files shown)


darothen (approved):
LGTM; much cleaner following our in-person review this morning!

There's a small nit here which is beyond the scope of this PR: how do make make the custom `_compute_metric` docstring appear as or override the docstring for `compute` for a given subclass? There's a cool trick to that which I'll implement in a follow-up.

PR #140, opened 2025-09-02
Atmospheric River Events, Metrics, Inputs

aaTman:
# EWB Pull Request

## Description

Draft atmospheric river pr, including IVT, mask, land intersection, etc. code.

PR #215, opened 2025-11-04
Fast mixed layer cape calculation (1/4)

darothen:
# EWB Pull Request

**NOTE** - This is the first in a sequence of commits that implements a fast mixed layer CAPE calculation based on MetPy's implementation. These PRs will merge to `feat/fast-cape` before we then merge that branch into `main`.

## Description

This PR implements a fast method for computing CAPE, using `numba`. 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

The numerical calculations here have a set of unit tests that capture known pathological profiles. We further produce a set of profiles curated from ERA5, and use correlations against the MetPy reference implementation to ensure that our method is behaving as expected.

## Checklist:

- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/extremeweatherbench/_cape.py
+++ b/src/extremeweatherbench/_cape.py
@@ -15,6 +15,7 @@ find it easier to use the xarray-based interfaces in the `cape.py` module.
 """
 
 import numpy as np
+import numpy.typing as npt
 from numba import njit, prange
 
 # ============================================================================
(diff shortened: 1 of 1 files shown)


src/extremeweatherbench/_cape.py
```diff
@@ -0,0 +1,724 @@
...
+            dewpoint_batch[i],
+            geopotential_batch[i],
+            depth,
+        )
+
+    return cape_array, cin_array
+
+
+def compute_ml_cape_cin_batched(
+    pressure: np.ndarray,
```
aaTman: Should return to this in the future to update typing

src/extremeweatherbench/_cape.py
```diff
@@ -0,0 +1,724 @@
...
+                z2 = heights[i + 1]
+
+            dz = z2 - z1
+            tv_avg = (env_tv[i] + env_tv[i + 1]) * 0.5
+            parcel_avg = (parcel_tv[i] + parcel_tv[i + 1]) * 0.5
+            energy = compute_buoyancy_energy_inline(parcel_avg, tv_avg, dz)
+            if energy > 0:
+                cape += energy
+
+    return cape, cin
```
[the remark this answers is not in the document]
darothen: My issue with a logger warning is that it may lead to users suppressing other userful warnings from the library. In this case I think we should just keep the docstring warning. I want to keep the interface so the method can be improved in the future more cleanly.

src/extremeweatherbench/_cape.py
darothen: Let's resolve it now - is there similar code elsewhere in the library you can point me towards? I'll adopt the convention of what's in the library currently.

aaTman: [Here in the atmospheric river PR](https://github.com/brightbandtech/ExtremeWeatherBench/pull/140/files#diff-07ad9cde8ee9c7e87f11805df88eca16252c1c68efc8c51ac31580c0bd0e305eR41) I'm using npt.NDArray.

aaTman (approved):
Once we swap the ndarray types, LGTM.

PR #228, opened 2025-11-12
restrict `metric_list` typing

aaTman:
# EWB Pull Request

## Description

In previous iterations of pre-release, there was an option to include a function that handles threshold-based metrics. The upcoming v1.0 release removes this and instead handles it entirely within either a `BaseMetric` or a child class of it. 

These typing changes reflect that and should not affect computations in the current `main` branch.

### DIFF
--- a/src/extremeweatherbench/evaluate.py
+++ b/src/extremeweatherbench/evaluate.py
@@ -11,11 +11,11 @@ import xarray as xr
 from tqdm.auto import tqdm
 from tqdm.contrib.logging import logging_redirect_tqdm
 
-from extremeweatherbench import cases, derived, inputs, utils
+from extremeweatherbench import cases, derived, inputs, metrics, utils
 from extremeweatherbench.defaults import OUTPUT_COLUMNS
 
 if TYPE_CHECKING:
-    from extremeweatherbench import metrics, regions
+    from extremeweatherbench import regions
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


src/extremeweatherbench/evaluate.py
```diff
@@ -174,8 +176,23 @@ def compute_case_operator(
         kwargs: Keyword arguments to pass to the metric computations.
 
     Returns:
-        A concatenated dataframe of the results of the case operator.
+        A concatenated dataframe of results from the case operator.
```
darothen: nit: a good convention is to refer to the object types explicitly in docstrings, to avoid ambiguity: "A concatenated pd.DataFrame of results from the CaseOperator".

src/extremeweatherbench/evaluate.py
```diff
@@ -174,8 +176,23 @@ def compute_case_operator(
...
+
+    Raises:
+        TypeError: If any metric is not properly instantiated (i.e. isn't an instance
+        or child class of BaseMetric).
     """
+    # Validate that all metrics are instantiated (not classes or callables)
+    for i, metric in enumerate(case_operator.metric_list):
+        if isinstance(metric, type):
+            metric = metric()
+            case_operator.metric_list[i] = metric
```
darothen: Would probably just do this in one line.

```suggestion
            case_operator.metric_list[i] = metric()
```

PR #234, opened 2025-11-19
Duration ME metric

aaTman:
# EWB Pull Request

## Description

This PR adds duration mean error to EWB. Duration mean error utilizes a climatology to determine the number of timesteps a case is beyond a climatological threshold for between a case's start and end date. 

Key improvements here include the use of a climatology as a reference for the forecast data, augmenting some of the sparse utility functions. 

OnsetME is temporarily removed for a forthcoming PR which will handle it separately.

`DurationME` also includes a `criteria` arg which allows a user to include a climatology or a float value alongside an `op_func` arg to provide an `operator` such as `operator.ge`.

### DIFF
--- a/src/extremeweatherbench/metrics.py
+++ b/src/extremeweatherbench/metrics.py
@@ -319,12 +319,19 @@ class ThresholdMetric(CompositeMetric):
     ) -> scores.categorical.BasicContingencyManager:
         """Create and transform a contingency manager.
 
+        This method is used to create and transform a contingency manager from the
+        scores module. The op_func is used to binarize the forecast and target data with
+        either a string representation of the operator, e.g. ">=", or a callable
+        function from the operator module, e.g. operator.ge.
+
         Args:
             forecast: The forecast DataArray.
             target: The target DataArray.
             forecast_threshold: Threshold for binarizing forecast.
             target_threshold: Threshold for binarizing target.
             preserve_dims: Dimension(s) to preserve during transform.
+            op_func: Function or string representation of the operator to apply to the
+            forecast and target. Defaults to operator.ge (greater than or equal to).
 
         Returns:
             Transformed contingency manager.


docs/examples/applied_freeze.py
```diff
@@ -71,12 +73,21 @@ def _preprocess_bb_cira_forecast_dataset(ds: xr.Dataset) -> xr.Dataset:
...
+
+# Define the metrics
 metrics_list = [
-    metrics.RMSE(),
-    metrics.MinimumMAE(),
-    metrics.OnsetME(),
-    metrics.DurationME(),
+    metrics.RMSE,
+    metrics.MinimumMAE,
+    metrics.DurationME(criteria=climatology, op_func=operator.le),
```
darothen: This is a little unergonomic for average scientific Python users. Let's add a utility somewhere that simplify defines a dictionary mapping the supported operators to their binary funcs, e.g.

``` python
operators = {
    ">": operator.gt,
    ">=": operator.ge,
    "<": operator.lt,
    "<=": operator.le
}
```

docs/examples/applied_heatwave.py
```diff
@@ -33,11 +35,19 @@
     variable_mapping=inputs.HRES_metadata_variable_mapping,
 )
 
+# Load the climatology for DurationME
+climatology = xr.open_zarr(
+    "gs://extremeweatherbench/datasets/surface_air_temperature_1990_2019_climatology.zarr",  # noqa: E501
+    storage_options={"anon": True},
+    chunks="auto",
+)
+climatology = climatology["2m_temperature"].sel(quantile=0.85)
```
darothen: Naming her is a tad confusing since you pass it to a kwarg "criteria" later on. Maybe rename this something like "temp_threshold" for clarity.

PR #243, opened 2025-11-21
Add default variable name reference

aaTman:
# EWB Pull Request

## Description

This PR simply adds a list of variable names in use inside EWB. The intent for these is to provide one single source of truth for mapping regardless of forecast or target source.


darothen (approved):
LGTM pending adding units to variable comments

PR #292, opened 2026-01-07
Improve and fix API dimension handling 

aaTman:
# EWB Pull Request

## Description

Although there was initially an option to include custom `preserve_dims` arguments, some of the metrics were unable to utilize it due to hardcoded reductions. This PR addresses this bug by moving many of the operations into distinctive functions in `utils.py` and ensuring that most reductions occur through the `preserve_dims` argument leading to `scores` infrastructure.

Some notes:

* `ForecastBase` objects now will not be coerced into `(lead_time, valid_time)` but stay as `(init_time, lead_time)`. This change produces some downstream consequences; namely, anything using `valid_time` directly now either must have been changed or modified to calculate `valid_time` when needed. This primarily impacts metrics behind-the-scenes but most are addressed here (except `DurationMeanError` which will be the subject of another PR).

* Minimum and Maximum MAE become very, very expensive if spatial dimensions aren't reduced off the bat, thus the default is to reduce latitude and longitude. I updated this argument to be Optional, such that a user can pass through None in the call (e.g. `metrics.MinimumMeanAbsoluteError(reduce_spatial_dims=None)` can be set as desired). I still don't recommend it unless willing to manage the computational *and* statistical nuances of point-by-point max and min handling.

* `derive_indices_from_init_time_and_lead_time` now takes two DataArrays instead of a Dataset with an implicit set of init and lead times.

* This argument pattern is continued for both new functions; that is, an init and lead time DataArray are required for computation.

## Type of change

## How Has This Been Tested?

All pytests pass, new tests were created and checked for expected inputs and behavior.

## Checklist:

### DIFF
--- a/src/extremeweatherbench/metrics.py
+++ b/src/extremeweatherbench/metrics.py
@@ -1047,6 +1047,7 @@ class MaximumMeanAbsoluteError(MeanAbsoluteError):
     def __init__(
         self,
         tolerance_range_hours: int = 24,
+        # TODO(293): Determine if this pattern is the best approach
         reduce_spatial_dims: list[str] = ["latitude", "longitude"],
         name: str = "MaximumMeanAbsoluteError",
         *args,


src/extremeweatherbench/metrics.py
```diff
@@ -1203,10 +1232,12 @@ def __init__(
         self,
         tolerance_range_hours: int = 24,
         name: str = "MaximumLowestMeanAbsoluteError",
+        reduce_spatial_dims: Optional[list[str]] = ["latitude", "longitude"],
```
darothen: Same comment as above.

src/extremeweatherbench/utils.py
```diff
@@ -183,15 +183,17 @@ def read_event_yaml(input_pth: str | pathlib.Path) -> dict:
 
 
 def derive_indices_from_init_time_and_lead_time(
-    dataset: xr.Dataset,
+    init_time_values: xr.DataArray,
+    lead_time_values: xr.DataArray,
```
darothen: From reading the function I'm not convinced these need to be `xr.DataArray`. I think any 1D array or vector would work here. This doesn't break because of the dispatching annotations on the `np.meshgrid` call you make in the function. I think the right type hint should be `numpy.typing.ArrayLike` (or `npt.ArrayLike` if you `import numpy.typing as npt` by convention).

Otherwise, you could revise L224 to pass in `lead_time_values.values` and `init_time_values.values` to `np.meshgrid` - possibly after checking that both arrays are actually 1D. And they probably should just be named `lead_times` and `init_times` to be cleaner.

src/extremeweatherbench/utils.py
```diff
@@ -864,3 +868,109 @@ def maybe_cache_and_compute(
...
+    valid_combinations_mask_da = xr.DataArray(
+        valid_combinations_mask,
+        dims=["init_time", "lead_time"],
+        coords={"init_time": init_time_values, "lead_time": lead_time_values},
+    )
+    return valid_combinations_mask_da
+
+
+def create_complete_day_mask(
+    init_time: xr.DataArray,
```
darothen: Same comment as above. Particularly relevant because you're constructing new xr.DataArrays as the output of this function, and you're not leaning on any specific DataArray capabilities as far as I can tell when you manipulate these vectors.

src/extremeweatherbench/metrics.py
```diff
@@ -1072,34 +1072,47 @@ def _compute_metric(
...
         """
-        # Enforced spatial reduction for MaximumMeanAbsoluteError
-        reduce_spatial_dims = ["latitude", "longitude"]
-        target_spatial_mean = utils.reduce_dataarray(
-            target, method="mean", reduce_dims=reduce_spatial_dims, skipna=True
-        )
-        maximum_timestep = target_spatial_mean.idxmax("valid_time")
-        maximum_value = target_spatial_mean.sel(valid_time=maximum_timestep)
+        # Reduce spatial dimensions if specified (default is to reduce)
+        if self.reduce_spatial_dims is not None:
```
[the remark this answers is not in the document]
aaTman: Ha, good to know! I think I knew this but definitely slipped my mind getting this set up. Surprising that none of the linters and type checkers caught this...

src/extremeweatherbench/metrics.py
```diff
@@ -1047,7 +1047,7 @@ class MaximumMeanAbsoluteError(MeanAbsoluteError):
     def __init__(
         self,
         tolerance_range_hours: int = 24,
-        reduce_spatial_dims: Optional[list[str]] = ["latitude", "longitude"],
+        reduce_spatial_dims: list[str] = ["latitude", "longitude"],
```
darothen: Sorry, I'm gonna cause a bit of whiplash here!

I think the right pattern should be to set `reduce_spatial_dims: Optional[list[str]] = None` in the signature, and then check `if reduce_spatial_dims is not None:` as the logical for doing the reduction. 

This means that by default, we _don't_ reduce the spatial dims (just temporal), and users must explicitly over-ride with a list of dimension names if they want to perform the reduction. I think this is the most consistent and predictable pattern, even if it does mean the caller needs to frequently pass more information. Explicit is better than implicit...

darothen (changes requested):
Thanks for the fast iteration! I apologize for the extra bit of work but see the new comment about `reduce_spatial_dims`... I think the approach I suggest in that comment is the best way to go, and should be very quick to refactor.

src/extremeweatherbench/metrics.py
darothen: Hmm, I think I understand what you mean. Then let's file an issue documenting this design conversation and add it to a future "TO-DO" list. I'll approve the PR pending a comment linking the issue here.

PR #367, opened 2026-04-24
fix: seed sample_calc_dataset fixture to eliminate flaky pressure test

darothen:
## Summary
- Seeds the `sample_calc_dataset` pytest fixture with `np.random.default_rng(42)` to produce bit-reproducible data across Python versions
- Fixes a flaky `test_pressure_at_surface` failure on Python 3.13 where unseeded random draws for orographic heights could produce extreme negative values, pushing barometric-formula surface pressure above the 105,000 Pa test bound

## Root cause
The fixture used the global numpy RNG (`np.random.normal`, `np.random.uniform`) without seeding. Python 3.13 initializes the global RNG state differently, allowing draws like `−320 m` for surface heights that are physically impossible but statistically valid — causing the derived surface pressure to exceed the test assertion bound.

## Test plan
- [ ] Run `pytest tests/test_calc.py::TestPressureCalculations::test_pressure_at_surface` — should pass consistently
- [ ] Run full test suite — no other test should be affected by the fixture seed change

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -391,7 +391,14 @@ def sample_calc_dataset():
     lon = np.linspace(-120, -80, 21)
     level = [1000, 850, 700, 500, 300, 200]
 
-    # Create realistic meteorological data
+    # Create realistic meteorological data.
+    # Use a seeded generator so results are bit-reproducible across Python
+    # versions; the global numpy RNG state can differ between interpreters and
+    # caused flaky failures when extreme draws produced physically impossible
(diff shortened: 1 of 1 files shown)


darothen:
This arises from an issue I encountered in #366 where we had flaky tests on Python 3.13 that I traced back to some changes in random number generation. The fix was extremely straightforward so I delegated to an agent. Will merge after checks pass.

PR #379, opened 2026-08-06
feat: opportunistically register dask-array for query-optimized xarray

aaTman:
## Summary

Integrates the [`dask-array`](https://github.com/mrocklin/dask-array) package (query-optimized reimplementation of `dask.array`) as a drop-in xarray chunk manager for EWB's evaluation pipeline, **on by default**.

- Adds `dask-array` as a core dependency in `pyproject.toml`, so a plain `pip install extremeweatherbench` gets it — this is opt-out, not opt-in.
- `evaluate.compute_case_operator()` registers it (`dask_array.xarray.register()`) once per process, before a case operator's target/forecast pipelines open any data. This is the earliest point in the codebase that is guaranteed to run in whichever process actually creates chunked arrays, whether that's the main process (serial mode), a `joblib` `loky` worker (parallel mode, default), or a `dask`-distributed worker — registering only in `ExtremeWeatherBench.run()`/`run_evaluation()` would not reach those worker subprocesses.
- Can be disabled via `evaluate.USE_DASK_ARRAY_QUERY_OPTIMIZATION = False` to fall back to the standard `dask.array` chunk manager. Also falls back gracefully (with a debug log) if `dask-array` is somehow not importable, e.g. a constrained environment installed without it.
- Documented in `docs/parallelism.md`, including the chunk-manager mixing hazard and how EWB's own known instance of it is handled (see below).

## Why here, not somewhere else

Two open PRs were checked first: [#377](https://github.com/brightbandtech/ExtremeWeatherBench/pull/377) (xarray/dask perf) and [#378](https://github.com/brightbandtech/ExtremeWeatherBench/pull/378) (CAPE fix, stacked on #377). Neither touches `evaluate.py`, so this is independent and low-conflict.

Chunked arrays are created lazily in `inputs.py` (`xr.open_zarr`/`open_dataset(chunks=...)`), but only once the pipeline actually runs — not at object-construction time. That run only happens inside `compute_case_operator()`, reached via `_build_datasets()` → `run_pipeline()`. `_maybe_register_optimized_dask_arrays()` is called at the very top of `compute_case_operator()` for that reason.

## The chunk-manager mixing hazard (and how it's handled)

Registering a chunk manager only affects arrays created afterward: xarray raises `TypeError: Mixing chunked array types is not supported` the moment an array created *before* registration is combined with one created *after* it, in the same process.

This PR went through a few iterations because of that:

1. First iteration shipped with this **on by default**, but a real evaluation run crashed with that exact `TypeError`. Root cause: `defaults.get_climatology()`, used by `defaults.get_brightband_evaluation_objects()` to build `DurationMeanError(threshold_criteria=...)`, opens a chunked, global-extent (~85 GB across all quantiles) `DataArray` in the main process before any `CaseOperator` runs — before registration ever fires. Worse, the crash wasn't just from combining it with forecast/target data directly: `climatology.interp_like(target, ...)` internally builds *new* interpolation-indexer arrays using whichever chunk manager is currently registered, then tries to combine those with climatology's own (differently-registered) array — so even operating on climatology alone, post-registration, could crash.
2. Second iteration reverted the default to opt-in (safe, but no longer matched the original ask) while the root cause was investigated.
3. Third iteration fixed the crash by having `utils.interp_climatology_to_target()` crop climatology to the case's time window and `.load()` that narrow slice into memory *before* interpolating, restoring the opt-out default. This fixed the crash but introduced a real performance regression: a real evaluation run (`docs/examples/applied_heatwave.py`) took **twice as long** on this branch as on `develop`. `.load()` forces a separate, blocking, synchronous network fetch on *every single case* (measured at ~20s per call against the real GCS climatology store), serializing work that used to be fused into one lazy graph alongside the forecast/target fetches.
4. This final iteration fixes that regression: `interp_climatology_to_target()` now calls `.chunk()` on climatology instead of `.load()`. Re-chunking an already-chunked array rewraps it onto whichever chunk manager is currently registered *without moving any data* — a graph-level operation that took ~0.1s even against the full ~85 GB global store in testing — so it stays fully lazy and avoids the mixing crash without the eager-fetch cost. End-to-end, a synthetic `DurationMeanError` call against the real store went from ~20s+ (eager `.load()`) down to ~2.4s (lazy `.chunk()`), consistent with pre-`dask-array` timing. Verified against the real GCS-hosted climatology dataset (`gs://extremeweatherbench/datasets/surface_air_temperature_1990_2019_climatology.zarr`, 85 GB / 12 GB per quantile) with a reproduction of the original crash, confirming both that it crashes without the fix and succeeds (and stays lazy) with it.

**Residual limitation:** this only covers EWB's own built-in usage. If you construct your own chunked array ahead of time and pass it into EWB (custom metric weights/thresholds, a dataset you already opened/chunked yourself before `XarrayForecast(ds=...)`, etc.), that array keeps using whichever chunk manager was active when you created it and can still crash if combined with a case operator's post-registration data. Documented in `docs/parallelism.md`, with `USE_DASK_ARRAY_QUERY_OPTIMIZATION = False` as the escape hatch.

## Benchmark (ad hoc, not part of the diff)

Measured with a throwaway script (not committed) that ran an EWB-shaped pipeline — `inputs.align_forecast_to_target`, a derived-variable-style elementwise op, a case-window `isel` subset, and `metrics.MeanAbsoluteError().compute_metric(...)` — against a synthetic dask-backed dataset, once with the standard `dask.array` chunk manager and once with `dask-array` registered. 3 fresh-process trials per configuration:

| Metric | Standard `dask.array` | `dask-array` registered | Change |
| --- | --- | --- | --- |
| Raw graph task count | 3054 | 204 | **15x fewer tasks** |
| Graph construction time (median) | ~0.32s | ~0.036s | **~9x faster** |
| `.compute()` wall time (median) | ~0.39s | ~0.054s | **~7x faster** |
| Peak RSS | ~453 MB | ~470 MB | **~4% higher** |

Speed and graph construction both improve substantially for this workload shape. Peak memory is a modest regression (~4%) at this scale, likely import/bookkeeping overhead for the new package rather than anything algorithmic — flagging it honestly rather than omitting it. (The plan's other proposed metric, task count after generic `dask.optimize(...)`, turned out not to be a fair comparison: `dask-array` returns a different, expression-based collection type that optimizes internally at `.compute()` time, and running the generic `dask.optimize` over it actually *increased* its task count from 204 to 260. Raw task count is the metric that's actually comparable across both chunk managers.)

## Test plan

- [x] `ruff check` / `ruff format --check` clean on all changed files.
- [x] `pytest tests/`: **1032 passed**. Includes:
  - `tests/test_evaluate.py`: registration when installed, graceful no-op when not importable, once-per-process idempotency, the `USE_DASK_ARRAY_QUERY_OPTIMIZATION` default value, and that `compute_case_operator` invokes the helper.
  - `tests/test_utils.py`: a regression test reproducing the original `TypeError` end-to-end with the real `dask-array` package (chunk a climatology-shaped array with the standard manager, register `dask-array`, confirm `interp_climatology_to_target` no longer crashes, stays lazily chunked under the new manager, and returns correct values).
  - `tests/conftest.py`: added an autouse fixture that reverts any real `dask-array` chunk-manager registration after each test — registration has no "unregister," so without this, the first test to exercise `compute_case_operator()` for real would permanently switch every later test in the session off the standard `dask.array` chunk manager.
- [x] Verified the climatology fix against the real GCS-hosted climatology store (not just synthetic data): confirms the re-chunk stays lazy (~0.1s, no data fetch) and the full pipeline (rewrap → interpolate → compute) takes ~2.4s end-to-end, versus 20s+ with the earlier eager-load approach.
- [x] `uv lock` regenerated cleanly after moving `dask-array` into core dependencies (net +57/-21 lines, scoped to the new package entry and its own dependency resolution).

Made with [Cursor](https://cursor.com)


aaTman:
## Closing: `dask-array` regresses EWB's actual workload

After the crash fix and the eager-load performance fix, a real `applied_heatwave.py` run still took over an hour on this branch vs. 42 minutes on `develop`. Root-caused it to something deeper than climatology: `inputs.align_forecast_to_target()` runs on *every* case operator and calls `forecast.interp(latitude=..., longitude=..., method="nearest", ...)` to regrid the forecast onto the target grid.

Isolated that exact call against the real HRES forecast store (`gs://weatherbench2/datasets/hres/2016-2022-0012-1440x721.zarr`, 210,494 chunks, 874 GB) with a case-sized time slice (`time=42, prediction_timedelta=41, latitude=721, longitude=1440`):

| Chunk manager | `.interp()` graph build time |
| --- | --- |
| Standard `dask.array` | ~0.03s |
| `dask-array` registered | ~0.72–0.86s |

Reproducible across repeated runs (not a one-off warm-up cost) — roughly a **25x slowdown just to build the graph**, before any compute. `dask-array`'s query optimizer is faster for some simpler patterns (plain time-slicing, small in-memory 2D interpolation), but for EWB's actual dominant pattern — regridding a large, many-chunked, lead-time-broadcast forecast onto a target grid, which runs on every single case operator, plus a similar `interp_like()` call in the climatology-based metric — it's a net regression. Multiplied across ~92 case operators (46 heatwave cases × 2 targets) with limited parallelism, this plausibly accounts for the observed slowdown.

This isn't fixable with a narrow patch; it's the optimizer's behavior on this specific (and pervasive) operation shape in an early-stage third-party package. Closing rather than merging a change that measurably regresses real evaluation runs.

